PR #6, opened 2015-04-21
Launch view to enable native iPhone6(+) resolution

jaanussiim:
Added just a empty xib. I suggest you add something to it. Maybe app logo?


wbaumann:
Yeah let's do the app image. What asset should I send you?

PR #7, opened 2015-04-22
Price object

jaanussiim:
Created WBPrice object and every price/tax field is converted to it. WBPrice holds amount and currency. Also money formatting is done there

### DIFF
--- a/SmartReceipts/WBPrice.m
+++ b/SmartReceipts/WBPrice.m
@@ -29,6 +29,18 @@
     return [WBPrice priceWithAmount:[NSDecimalNumber zero] currencyCode:currencyCode];
 }
 
+static NSNumberFormatter *__noCurrencyFormatter;
+- (NSString *)amountAsString {
+    if (!__noCurrencyFormatter) {
+        __noCurrencyFormatter = [[NSNumberFormatter alloc] init];
+        [__noCurrencyFormatter setNumberStyle:NSNumberFormatterCurrencyStyle];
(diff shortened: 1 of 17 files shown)


SmartReceipts/WBPrice.m
```diff
@@ -0,0 +1,71 @@
...
+        formatter = [[NSNumberFormatter alloc] init];
+        [formatter setNumberStyle:NSNumberFormatterCurrencyStyle];
+        [formatter setCurrencyCode:code];
+        [WBPrice sharedFormatterCache][code] = formatter;
+    }
+
+    return formatter;
+}
+
++ (NSMutableDictionary *)sharedFormatterCache {
```
wbaumann: I'm worried about possible formatting issues on this one, since NSNumberFormatter is not thread safe (https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html). The same is true of the Java Date Formatter equivalent, so I create a new one each time (although I could create a synchronized version as well).

I'm not sure if/how this would manifest itself in Obj-C, so let me know if this is a non issue. In Java at least though, calling format on different threads can potentially corrupt the results of both due to lack of thread safety

jaanussiim:
@wbaumann, you are right. Everything expect exporters would have been running in UI thread. I would still cache and moved these to [[NSThread currentThread] threadDictionary].

wbaumann:
Sounds good. Thanks!

PR #8, opened 2015-04-22
Extracted image picker to separate class

jaanussiim:
Extracted image picker presentation to separate class. This enabled single place to handle picker callbacks (needed to reset status bar color). 

This also allowed me to remove some copy/paste code related to image picker delegate methods

### DIFF
--- a/SmartReceipts.xcodeproj/project.pbxproj
+++ b/SmartReceipts.xcodeproj/project.pbxproj
@@ -136,6 +136,7 @@
 		EA3CF0971AE77E1900B14B73 /* WBPrice.m in Sources */ = {isa = PBXBuildFile; fileRef = EA3CF0961AE77E1900B14B73 /* WBPrice.m */; };
 		EA3CF09B1AE7964900B14B73 /* NSDecimalNumber+WBNumberParse.m in Sources */ = {isa = PBXBuildFile; fileRef = EA3CF09A1AE7964900B14B73 /* NSDecimalNumber+WBNumberParse.m */; };
 		EA3CF09E1AE7C45C00B14B73 /* Constants.m in Sources */ = {isa = PBXBuildFile; fileRef = EA3CF09D1AE7C45C00B14B73 /* Constants.m */; };
+		EA46678F1AE7EA1700B4E502 /* WBImagePicker.m in Sources */ = {isa = PBXBuildFile; fileRef = EA46678E1AE7EA1600B4E502 /* WBImagePicker.m */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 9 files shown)


SmartReceipts.xcodeproj/project.pbxproj
```diff
@@ -136,6 +136,7 @@
 		EA3CF0971AE77E1900B14B73 /* WBPrice.m in Sources */ = {isa = PBXBuildFile; fileRef = EA3CF0961AE77E1900B14B73 /* WBPrice.m */; };
 		EA3CF09B1AE7964900B14B73 /* NSDecimalNumber+WBNumberParse.m in Sources */ = {isa = PBXBuildFile; fileRef = EA3CF09A1AE7964900B14B73 /* NSDecimalNumber+WBNumberParse.m */; };
 		EA3CF09E1AE7C45C00B14B73 /* Constants.m in Sources */ = {isa = PBXBuildFile; fileRef = EA3CF09D1AE7C45C00B14B73 /* Constants.m */; };
+		EA46678F1AE7EA1700B4E502 /* WBImagePicker.m in Sources */ = {isa = PBXBuildFile; fileRef = EA46678E1AE7EA1600B4E502 /* WBImagePicker.m */; };
```
wbaumann: Since this file seems to keep popping up, do you think it makes sense to add this to the gitignore file? This are purely just project level changes, correct?

SmartReceipts/Utils/WBImagePicker.m
```diff
@@ -0,0 +1,106 @@
...
+
+@interface WBImagePicker () <UINavigationControllerDelegate, UIImagePickerControllerDelegate>
+
+@property (nonatomic, copy) WBImagePickerResultBlock selectionHandler;
+
+@end
+
+@implementation WBImagePicker
+
++ (instancetype)sharedInstance {
```
wbaumann: Does this need to be a singleton? And what will it buy us in this case?

The only hesitation with this situation is that it will make unit tests more difficult, since you cannot mock singletons

wbaumann:
Looks good overall - I just want to check about the singletons, since it may make future unit tests more difficult. In general, I'm completely fine with the slight overhead of alloc init to enable future testing ease. We can discuss offline if you're unfamiliar with mocks, since I can walk through some of the benefits there.

SmartReceipts.xcodeproj/project.pbxproj
jaanussiim: No, project.pbxproj can't be added to gitignore. This is the file that tells Xcode where everything is, how it shown in Xcode and how app is built.

SmartReceipts/Utils/WBImagePicker.m
jaanussiim: Intent with sharedInstance was not to reduce alloc/init calls, but rather simplify calling picker code. UIImagePickerController delegate (WBImagePicker) needs to stick around until picker has complete. For that there are two options - manage WBImagePicker instance where ever it is called. Or manage WBImagePicker through shared instance. 

Opted for second option because of cleaner code on the calling side. Just invoke WBImagePicker and wait for response, without any additional memory management.

jaanussiim:
I know what mocks are and how they are used.

wbaumann:
Cool - all sounds good and makes sense. Thanks for the explanation

PR #11, opened 2015-04-24
Implement receipt colums

jaanussiim:
Implemented columns and converted csv/pdf creation to that. This is a straight column resolver replacement. Could need some cleanup and simplification.

This pull is not finished yet. As mentioned - some more cleanup is neede. Just wanted to show the work in progress and get confirmation it's in right direction :)


SmartReceiptsTests/Reports/WBReportUtilsTest.m
```diff
@@ -0,0 +1,54 @@
...
+//
+
+#import <UIKit/UIKit.h>
+#import <XCTest/XCTest.h>
+#import "WBPreferences.h"
+#import "WBReceipt.h"
+#import "WBReportUtils.h"
+#import "WBPrice.h"
+
+@interface WBReportUtilsTest : XCTestCase
```
wbaumann: Nice :)

wbaumann:
Yep - looks great! Thanks for putting this together

jaanussiim:
Now I should be finished with this

PR #13, opened 2015-04-26
Table generator

jaanussiim:
Added generic table generators that take columns and rows as input.


SmartReceipts/Reports/Tables/ReportCSVTable.m
```diff
@@ -0,0 +1,85 @@
...
+
+static NSString* const QUOTE = @"\"";
+static NSString* const ESCAPED_QUOTE = @"\"\"";
+
+static NSString *escapedCSV(NSString* csv) {
+    if (!csv) {
+        return @"";
+    }
+
+#warning FIXME: on Android is csv.replace(QUOTE, ESCAPED_QUOTE); that means that result of replacement is not saved
```
wbaumann: FYI - This line is no longer relevant. We can delete as part of another pull

PR #14, opened 2015-04-26
Warnings

jaanussiim:
This pull contain some project cleanup. Turned on some more warnings and moved most of thirdparty code to cocoapods. Nothing really to review here. 

The ones that were moved were just simple switch. Two were not moved, as these needed some code changes and did not want to waste time on that.

Could you also go over #warning pragmas that are turning up under build warnings? Remove the ones that are not relevant or create Trello issues for the ones that need some work.

Some stupid mistakes can be easily discovered during project compiling. But only if there is no other noise. My goal would be to get warnings to zero.


wbaumann:
Sounds good. I'll do a pass through the code base, remove all irrelevant pragmas, and open trello tickets for the rest. Thanks!

PR #15, opened 2015-04-29
Migrate to v13

jaanussiim:
Well... This was fun :)

Created migrations to have parity with Android. Untangled some database code.

When you provide me DB file from Android, I will add additional migration checks - will verify that iOS migration to v13 will result in database that has same tables and columns.


wbaumann:
Awesome - looks great. I'll get you samples from database versions 10, 11, and 12 for additional test verification

PR #16, opened 2015-05-05
Mileage entry/modify

jaanussiim:
This contains also bigger implementation ideas:

1) how to create and show controllers with list of input cells (EditDistanceViewController)
2) how to present list rows from DB (TripDistancesViewController and FetchedCollectionTableViewController)
3) some ideas how to do object saves and updates based on that to UI (FetchedModelAdapter)


SmartReceipts/Model/WBTrip.m
```diff
@@ -10,7 +10,7 @@
 #import "WBFileManager.h"
 #import "WBPrice.h"
 
-static NSString * const MULTI_CURRENCY = @"XXXXXX";
+NSString *const MULTI_CURRENCY = @"XXXXXX";
```
wbaumann: As a heads up, this is a legacy implementation detail that will hopefully be going away soon. On the Android side, I have one Price.java implementation that handles this more properly

PR #21, opened 2015-05-12
Mileage total in report setting

jaanussiim:
Tried to do my best to cover usage of this setting with unit tests. This meant also some work on generalising other parts of database code. There are some more cases that would ned unification and review. I see it as additional work for future tasks.

Because of database locking one pattern emerged where some methods got a suffix 'usingDatabase...'. I don't like it. Will see if I can get rid of it in the future.


wbaumann:
Thanks for some database generalizations. I've started a similar approach on Android but will probably look into adopting more of what you've done here.

As for the locking, I "think" we may be able to get rid of it entirely (but will have to review further). I believe that SQLite will automatically handle the locking for you so long as there's only a single instance of the database active. At least on the Android side, you're okay as long as you only have a single activ connection (http://kagii.tumblr.com/post/6828016869/android-sqlite-locking). I'm not sure if it's done at the Android level or SQLite level though, so it's hard to say if this will translate to iOS

PR #38, opened 2015-06-01
Export import

jaanussiim:
Had some trouble figuring out how to test imports. Opted for simplified database import. Rules are as follows:

1) Overwrite CSV and PDF columns. Could be annoying, but no data lost.
2) Import added values for payment methods and categories. This ensures that nothing is lost. Deletions in imported database will not be imported.
3) For trips handle every trip object as whole. Receipts and distances are considered part of it. And overwrite rule is used as follows - when yes and have existing trip with same name, delete that one and import new one. When no and have existing one ignore it and go to next trip import.

Note that there is no separate receipts or distances checking. Everything is part of trip import.

Also opted to use all the methods already in DB interface, to keep the code simpler.


wbaumann:
Looks good. I'll test out with a bunch of SMR files that I have already and will let you know if I hit any issues

PR #58, opened 2015-06-26
Trip price update

jaanussiim:
On 'OnlyIncludeExpensable' and 'MileageTotalInReport' change trip prices in database will be updated. 

Did I miss any more relevant settings?


wbaumann:
Think that should be good. Will do some more testing with the latest and let you know if we missed anything

PR #61, opened 2015-07-03
In app purchase

jaanussiim:
This should have everything in place for IAP-s.

Should we do subscription end verification in this release or will we remember to do it before year has passed? :)


wbaumann:
Yea - let's do it in this release, so we don't forget a year from now

PR #62, opened 2015-07-03
Cache entry values

jaanussiim:
Currently cache validity is tied to trip controller. When ever new trip is opened, cached data is cleared.


wbaumann:
Perfect - that's the same as Android

PR #70, opened 2015-08-15
Negative number entry

jaanussiim:
This pull contains demo of two options I see. 

We can use keyboard that allows minus sign entry. Or we could add button to keyboard that specifies if positive or negative number is entered.

When you enable tax field entry and open receipt edit screen, price field will use the first option. Tax field will use the second option.

Switching to first option adds no extra work. Switching to second option will add some additional logic to insert/delete minus sign to entry and not allowing user to delete it.

### DIFF
--- a/SmartReceipts/Cells/TextEntryCell.h
+++ b/SmartReceipts/Cells/TextEntryCell.h
@@ -21,5 +21,6 @@
 - (void)activateNumberEntryMode;
 - (void)activateEmailMode;
 - (void)setValue:(NSString *)value;
+- (void)addAccessoryViewWithNegativeSwitch;
 
 @end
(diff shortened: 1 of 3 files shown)


wbaumann:
Merging in for testing. What do you think about handling devices (e.g. iPads) that show the minus sign by default (due to having a larger keyboard)

jaanussiim:
iPad defaults to first option - keyboard allowing minus sign entry.

jaanussiim:
Or I guess iPad will stay the same as it is. Just need to decide what to do on phone side :)

wbaumann:
Tried both. Let's go with the first option. I think that'll be easier

PR #78, opened 2016-05-18
Fix some tests

jaanussiim:
This fixes some tests, not all.

DatabaseDatesTest tests is using receipts-dates-test.db that has been generated using some older version of SmartReceipts. It contains receipts that have no paymentmethodkey defined and retrieving receipts fails. To get around it I added partial v14 migration to set defaults there. So I believe for data consistency we do need to include this defaulting.

Updated payment methods import to latest implementation that does overwrite.

Found and fixed issue with categories import. When override is used, then you may end up with no categories, as local ones were removed, but during insert we were checking still against local deleted categories for duplicates.


wbaumann:
Looks good. Thanks for fixing some of the tests that I broke as well

PR #80, opened 2016-05-24
Start exchange rate

jaanussiim:
Started with small UI update to allow exchange rates entry.

Most of work was checking v14 migration. 

For payment method mapping I suggest to use LEFT JOIN. Original idea to default it during v14 migration will not work when importing DB that is already on v14 (from Android).

For current tests LEFT JOIN seems to be working (removed default payment method setting).

Current exchange rate support is minimum code to have import working. It's not used anywhere in UI.


SmartReceipts/Constants.m
```diff
@@ -22,6 +22,8 @@
 
 NSString *const SmartReceiptSubscriptionIAPIdentifier = @"co.smartreceipts.ios.pro.noads";
 
+NSString *const SRNoData = @"null";
```
wbaumann: Fwiw - I'm basically deprecating this on the Android side (in favor of actual DB nulls)

SmartReceipts/Persistence/Database.h
```diff
@@ -23,9 +23,14 @@ extern NSString *const DatabaseDidSwapModelsNotification;
...
-- (id)initWithDatabasePath:(NSString *)path tripsFolederPath:(NSString *)tripsFolderPath;
 - (BOOL)open;
-- (BOOL)open:(BOOL)migrateDatabase;
 - (void)close;
 
 @end
+
+@interface Database (ExposeForTests)
+
+- (id)initWithDatabasePath:(NSString *)path tripsFolederPath:(NSString *)tripsFolderPath;
```
wbaumann: typo in "tripsFolederPath" - I'll merge to unblock you but could you fix in the next one

PR #81, opened 2016-05-30
Exchange rate entry

jaanussiim:
Changed WBReceipt interface to get better understanding how/where exchange rate will be used. This change was also needed to have price, trip currency and exchange rate closer together for calculations.

Calculations are used in report columns, will need to investigate what to do in trips list.

One note - currently exchange rate and tax related formats (amountAsStrins, formattedAmount, etc) will return empty sting when these values are missing.

Meaning that user adds 'Price (Exchanged)' column, but currency does not require exchange calculations, empty column will be shown.


wbaumann:
Looks good. Let me double check the Android side of things to make sure we're consistent on the price (exchanged) item (i.e. blank vs original price).

PR #82, opened 2016-06-01
Trips price handling change

jaanussiim:
This changes trip price handling to use calculations during fetch. Previously price column in database was used and on every receipt/distance change updated.

Removed MULTY_CURRENCY references.

Introduced a query patter on top of FMDatabase.

For previous pattern we have it as follows:
Database -> FMDatabaseQuery -> FMDatabase

When performing queries using Database, it created pattern with two methods:
1) method
2) methodUsingDatabase:(FMDatabase *)database

I'm hoping that making queries on top of FMDatabase will make a better interface.


wbaumann:
Looks cool with the new query pattern. thanks!

PR #87, opened 2016-06-07
Send reports

jaanussiim:
This demonstrates two ways to do the 'Other' sharing: UIActivityViewController and UIDocumentInteractionController.

UIActivityViewController can take multiple files, but has fewer options.

UIDocumentInteractionController takes one file, but allows access to more applications and options.

Second one is definitely better, but that would require change in 'generate' UI to allow generation of only one file.


wbaumann:
Cool - I'll test these out. Quick question: what's your thought for the best approach here? I was thinking:

- If the user only has a single document, use the UIDocumentInteractionController flow
- If the user has multiple, use UIActivityViewController

I'd worry that reverting to only allowing a single file at a time may frustrate users that are familiar with this existing pattern...

jaanussiim:
UIActivityViewController and UIDocumentInteractionController have different options for sharing. UIDocumentInteractionController having more options. Choosing the sharing option based on number of files generated would crate a flow where with one file you can share to almost anywhere, but with multiple files limited sharing will be available.

PR #97, opened 2016-11-22
Swift 3 code migration

bananafish911:
tests OK

* I'm not sure about Tweaker.swift functionality (how does it should work? the code looks like mix of swift and obj-c syntaxes, so I left it "as is")


wbaumann:
Tweaker was a utility class to allow us to toggle purchases on/off without actually needing to register a test purchase account with Apple No worries about it though. Thanks for the rest!

PR #100, opened 2016-12-12
Setting VC - feedback and Assume Full Page

bananafish911:
Pushing already finished small changes: 
- added "Feedback" Settings
- added "Assume Full Page Receipt Image"


SmartReceipts/Constants.m
```diff
@@ -26,4 +26,8 @@
 
 void SRDelayedExecution(NSTimeInterval seconds, ActionBlock action) {
     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, seconds * NSEC_PER_SEC), dispatch_get_main_queue(), action);
-}
\ No newline at end of file
+}
+
+NSString *const FeedbackEmailAddress =              @"will.r.baumann@gmail.com";
```
wbaumann: Can you make this a 'concatenated' string instead? Something like:

[NSString stringWithFormat:@"%@/%@/%@", @"will.r.baum", "ann@gm", "ail.com"];

Not that it's a huge deal, but I'd prefer to reduce the likelihood of crawler grabbing my email

SmartReceipts/Extensions/UIDevice+DeviceInfo.m
```diff
@@ -0,0 +1,101 @@
...
+    uname(&systemInfo);
+    
+    NSString* code = [NSString stringWithCString:systemInfo.machine
+                                        encoding:NSUTF8StringEncoding];
+    
+    static NSDictionary* deviceNamesByCode = nil;
+    
+    if (!deviceNamesByCode) {
+        
+        deviceNamesByCode = @{@"i386"      :@"Simulator",
```
wbaumann: Was there a stack overflow article (or something) where you got these? If so could you add it in the comments, so I can update this in the future whenever iPhone 8 + comes out

PR #101, opened 2016-12-28
Squashed commit: feature-analytics

bananafish911:
Cocoapods were updated
SmartReceipts/ServiceAccount.json was added to .gitignore as a secret file


wbaumann:
Awesome. Looks great :)

PR #112, opened 2017-04-10
Develop

bananafish911:
Fixes: "Backup menu appears twice when launched from a cold start"

What's new: I've added database migration up to v15, unit tests and few Logger messages for error tracking when migrating.


wbaumann:
Looks great. Thanks Victor!

PR #113, opened 2017-04-17
Develop

bananafish911:
Few tasks from queue:
+ Rename "Trip" to "Report" on iOS
+ Add recently used currencies are at the top of the list (and unit tests)

### DIFF
--- a/SmartReceipts/ViewControllers/SettingsViewController.swift
+++ b/SmartReceipts/ViewControllers/SettingsViewController.swift
@@ -33,6 +33,9 @@ extension SettingsViewController {
     
     func analyticsSettingsOverflow() {
         AnalyticsManager.sharedManager.record(event: Event.Navigation.SettingsOverflow)
+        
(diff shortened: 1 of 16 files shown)


SmartReceipts/ViewControllers/SettingsViewController.swift
```diff
@@ -33,6 +33,9 @@ extension SettingsViewController {
     
     func analyticsSettingsOverflow() {
         AnalyticsManager.sharedManager.record(event: Event.Navigation.SettingsOverflow)
+        
+        // FIXME: remove
```
wbaumann: Should we remove this now?

bananafish911: Yep, my mistake. Will be removed in next pull request.

wbaumann: Thanks :)

PR #122, opened 2017-05-30
Files structure & subtitle

EvsenevDev:
- Added subtitle to Navigation Bar
- Changed: same paths for Folders like Groups in Project


wbaumann:
Thank you for this :). Looks great!

PR #123, opened 2017-06-03
VIPER, Eureka and Swift migration

EvsenevDev:
- I added pods Viperit and Eureka to project.
- Also i rewrote Fetched Collection Table View to Swift (but i not removed old class from project) and added it to new VIPER modules.
- Fixed placeholders for UIViewController.
- Rewritten WBCustomization to Customization.swift.
- Some classes rewitten to Swift.
- Removed TripDeistancesViewController and EditDistanceViewController

[DistanceEditor for compare](https://gist.github.com/EvsenevDev/ed5f3ac272950e9a2a2250fe330171c5)

**So, it's interesting PR**


SmartReceipts/Modules/Edit Distance/EditDistanceInteractor.swift
```diff
@@ -0,0 +1,30 @@
...
+//  Copyright © 2017 Will Baumann. All rights reserved.
+//
+
+import Foundation
+import Viperit
+
+final class EditDistanceInteractor: Interactor {
+    func save(distance: Distance, asNewDistance: Bool) {
+        if (asNewDistance) {
+            AnalyticsManager.sharedManager.record(event: Event.distancePersistNewDistance())
```
wbaumann: I forgot to mention this in our initial discussions (my fault), but let's also add log statements here to help with debugging. The main premise is that we should avoid PII data in our logs, so something like "Saving a new distance" or "Updating an existing distance" (ie not emitting the user's actual distance data).

SmartReceipts/Modules/Trip Distances/TripDistancesView.swift
```diff
@@ -0,0 +1,125 @@
...
+                summaryCell.totalLabel.text = distance.totalRate().mileageRateCurrencyFormattedPrice()
+                summaryCell.dateLabel.text = dateFormatter.string(from: distance.date)
+                summaryCell.setPriceLabelWidth(maxRateWidth)
+            }
+        }
+    }
+    
+    //MARK: Actions
+    private func configureUIActions() {
+        _ = doneButtonItem?.rx.tap.subscribe(onNext: { [weak self] in
```
wbaumann: Just double check that we don't need to add these to a `DisposeBag`. I don't think we do but it's useful for avoiding potential memory leaks

SmartReceipts/ViewControllers/Abstract/Customization/Customization.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Customization.swift
+//  SmartReceipts
+//
+//  Created by Bogdan Evsenev on 01/06/2017.
+//  Copyright © 2017 Will Baumann. All rights reserved.
+//
+
+import Foundation
+
+class Customization: NSObject {
```
wbaumann: Let's rename this to `Theme` or `AppTheme` to be clearer about it's use (bad naming from before)

SmartReceipts/Modules/Edit Distance/EditDistanceInteractor.swift
```diff
@@ -0,0 +1,30 @@
...
+//  SmartReceipts
+//
+//  Created by Bogdan Evsenev on 01/06/2017.
+//  Copyright © 2017 Will Baumann. All rights reserved.
+//
+
+import Foundation
+import Viperit
+
+final class EditDistanceInteractor: Interactor {
```
wbaumann: For each of these new interactor and presenter implementations, could you also add unit test coverage to make sure that we're testing the core business logic? Let me know if you're not super familiar with this, but we should already have a number of tests in the app.

For this class for example, you'd want to do the following:

- Add the `Database.sharedInstance` as a constructor param (ie to allow mocking... or at least to the degree swift allows: https://news.realm.io/news/tryswift-veronica-ray-real-world-mocking-swift/). 
- Verify if `save(Distance)` returns `true` then we close the presenter
- Verify if `save(Distance)` returns `false` then we don't close the presenter

Everyone has their own philosophy regarding testing, but mine is:

- They're a major annoyance to write
- But they're extremely helpful in a team project (ie I cannot accidentally break something that you created without also breaking the tests)

SmartReceipts/Modules/Trip Distances/TripDistancesView.swift
EvsenevDev: Yes, better is add UI Controls to a `DisposeBag`

SmartReceipts/Extensions/String+Extensions.swift
```diff
@@ -53,3 +53,12 @@ extension String {
         return string
     }
 }
+
+extension String {
+    mutating func appendIssue(_ issue: String) {
+        if !self.isEmpty {
+            self = self.appending("\n")
+        }
+        self = self.appending("• \(issue)")
```
EvsenevDev: ok

SmartReceipts/Modules/Edit Distance/EditDistanceInteractor.swift
EvsenevDev: Ok, i will add Unit Tests in next commit

SmartReceipts/ViewControllers/Abstract/Customization/Customization.swift
EvsenevDev: `AppTheme` sounds better ;)

SmartReceipts/Modules/Edit Distance/EditDistanceInteractor.swift
EvsenevDev: ok

wbaumann:
@EvsenevDev Merging now! We add the requested changes as part of the next PR. Thanks!

PR #125, opened 2017-06-09
Generate report module

EvsenevDev:
- Fixed typo "distanceSaved"
- Chenged to "Logger.error" log in nagative cases of call db trasnsaction
- Added "Generate Report" Module (VIPER + Rx)
- Removed old "Generate Report" screen
- Prepared "Generate Report" module mocks for Unit Tests
- Added Unit Tests for Generate Report Module
- Updated Unit Tests for Edit Distance & Trip Distances Modules


SmartReceipts/Modules/Generate Report/GenerateReportPresenter.swift
```diff
@@ -0,0 +1,75 @@
...
+    let csvFile = BehaviorSubject<Bool>(value: false)
+    let zipStampedJPGs = BehaviorSubject<Bool>(value: false)
+    
+    override func viewHasLoaded() {
+        interactor.configureBinding()
+        interactor.trackConfigureReportEvent()
+    }
+    
+    override func setupView(data: Any) {
+        if let trip = data as? WBTrip {
```
wbaumann: Is there any cases in which this wouldn't be a trip?

Let's force the case (let trip = data as! WBTrip) instead, so we crash if we ever mess this up

SmartReceiptsTests/Modules Tests/Generate Report/GenerateReportModuleTest.swift
```diff
@@ -0,0 +1,103 @@
...
+        verify(interactor).trackConfigureReportEvent()
+        
+        presenter.generateReport()
+        verify(interactor).trackGeneratorEvents()
+        verify(interactor).generateReport()
+    }
+    
+    func testPresenterToRouter() {
+        presenter.close()
+        verify(router).close()
```
wbaumann: Nice :)

PR #129, opened 2017-06-16
Fixed Gas Rate and No Trips for iPad

EvsenevDev:
- Fixed "Gas Rate" from Settings
- [iPad] Fixed NoTripsViewController
- Fixed Xcode warinings


wbaumann:
Looks great. Thanks :)

PR #136, opened 2017-07-16
Settings Module 

EvsenevDev:
Settings Module with VIPER and Rx


SmartReceipts/Eureka Custom/InputTextCell.swift
```diff
@@ -0,0 +1,52 @@
...
+    @IBOutlet weak var valueField: UITextField!
+    
+    private var textFieldColor: UIColor!
+    
+    public override func setup() {
+        textField = valueField
+        titleLabel = label
+        textFieldColor = valueField.backgroundColor
+        super.setup()
+        height = { 52 }
```
wbaumann: Can we move the `52` value to a `SettingsTheme#getSettingsRowHeight` or `AppTheme#getSettingsRowHeight` something, so we can change this from a single place? I see you've got it in a few already

PR #138, opened 2017-08-08
Prepare 1.6

EvsenevDev:
- Added a "Receipts Default As Reimbursable" to iOS
- Added logging for Report insert/update failures
- [Fix] Wrong initial Rate on create Distance
- [Fix] Files, images and attach it to receipts with omitted names
- [Fix] Compiler Warnings after Pods updated

And change some spaces in WBPreference for better code style.

### DIFF
--- a/SmartReceipts/Persistence/WBPreferences.m
+++ b/SmartReceipts/Persistence/WBPreferences.m
@@ -197,7 +197,7 @@ static NSDictionary *getDefaultValues() {
             BOOL_CAMERA_ROTATE_IMAGE : @YES,
 
             BOOL_INCLUDE_MILEAGE_TOTAL_IN_REPORT: @NO,
-            FLOAT_DEFAULT_MILEAGE_RATE: @(MIN_FLOAT),
+            FLOAT_DEFAULT_MILEAGE_RATE: @(0),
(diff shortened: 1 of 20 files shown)


SmartReceipts/Persistence/WBPreferences.m
```diff
@@ -194,7 +197,7 @@ typedef NS_ENUM(short, EntryType){
             BOOL_CAMERA_ROTATE_IMAGE : @YES,
 
             BOOL_INCLUDE_MILEAGE_TOTAL_IN_REPORT: @NO,
-            FLOAT_DEFAULT_MILEAGE_RATE: @(MIN_FLOAT),
+            FLOAT_DEFAULT_MILEAGE_RATE: @(0),
```
EvsenevDev: This default value for Distance Rate i think 0 is better and if 0 in Rate i check it and hide value in Edit Distance Module, i just show values > 0 and < 0 for. In past it was MIN_FLOAT in Default, but i don't know reason. Do you want to change it back or know reason of MIN_FLOAT?

wbaumann: Nope. That sounds good. Just wanted to confirm. Thanks :)

PR #145, opened 2017-09-01
Report tooltip

EvsenevDev:
- Added TooltipView which you can show on UIView
- Added TooltipService for trigger management and service have factory method for tooltip texts


SmartReceipts/Services/TooltipService.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+    
+    func markReportGenerated() {
+        mark(key: REPORT_GENERATED)
+    }
+    
+    // Get
+    func moveToGenerateTrigger() -> Bool {
+        let trips = database.allTrips() as! [WBTrip]
+        let justOneTrip = trips.count == 1
```
wbaumann: I think this is slightly different than the Android specs (so sorry for the confusion), but can we make this to be `> 0` as well, so we can guide all users to this? This won't be an issue for new installs (it'll still show on the first trip), but it'll help users that aren't familiar with the UI changes that we're introducing here

EvsenevDev: ok, we will change this condition, right?
 **The user only has a single report**

wbaumann:
Correct. Instead of just a single report, we'll allow any number greater
than 0. All other conditions should remain the same

PR #151, opened 2017-09-22
Cognito

EvsenevDev:
- Added **getUser** to AuthService
- Added **getUser** test in AuthService
- Updated pods for Xcode 9
- Prepared CognitoSerivce
- Added FeatureFlag - **UseProdEndpoints**

### DIFF
--- a/SmartReceipts/Model/User.swift
+++ b/SmartReceipts/Model/User.swift
@@ -1,0 +1,25 @@
+//
+//  User.swift
+//  SmartReceipts
+//
+//  Created by Bogdan Evsenev on 21/09/2017.
+//  Copyright © 2017 Will Baumann. All rights reserved.
+//
+
+import SwiftyJSON
+
+class User {
+    var cognitoToken: String!
+    var id: String!
+    var identityId: String!
+    var email: String!
+    
+    init(_ json: JSON) {
+        let user = json["user"]
+        cognitoToken = user["cognito_token"].stringValue
+        id = user["id"].stringValue
+        identityId = user["identity_id"].stringValue
+        email = user["email"].stringValue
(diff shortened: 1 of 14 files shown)


SmartReceipts/Model/User.swift
```diff
@@ -0,0 +1,25 @@
...
+//  SmartReceipts
+//
+//  Created by Bogdan Evsenev on 21/09/2017.
+//  Copyright © 2017 Will Baumann. All rights reserved.
+//
+
+import SwiftyJSON
+
+class User {
+    var cognitoToken: String!
```
wbaumann: While this should never happen, I think it's safer to make these all optional (ie String?), since there's never a guarantee of what our services will actually return (even though they are supposed to be returning all of these :) ). But if there's ever a backend bug, I wouldn't want it to crash the app or anything

SmartReceipts/Services/CognitoService.swift
```diff
@@ -0,0 +1,62 @@
...
+        AuthService.loggedInObservable
+            .filter({ $0 })
+            .flatMap({ [unowned self] _ in
+                return self.authService.getUser()
+            }).bind(to: userVar)
+            .disposed(by: bag)
+    }
+    
+    override var identityProviderName: String {
+        return "login.smartreceipts.co"
```
wbaumann: All this looks right :)

SmartReceipts/Model/User.swift
EvsenevDev: @wbaumann I will make Failable init for User, i.e. if cognito Token or identityId are empty User will be nil, because without Cognito credentials User is useless, what do you think?

PR #156, opened 2017-09-30
Service push notifications

EvsenevDev:
- **AppDelegate** rewritten to Swift
- Added **Push Notification Service**
- Added **saveDevice(token:)** to AuthService for upload FCM token

Push Notification Service:
- Handle Notifications and unwrap JSON then emit it into **notificationJSON** observable.
- Handle and Send FCM token to **users/me**


SmartReceipts/AppDelegate.swift
```diff
@@ -0,0 +1,194 @@
...
+        
+        Logger.info("Language: \(Locale.preferredLanguages.first!)")
+        
+        NSSetUncaughtExceptionHandler { exception in
+            RateApplication.sharedInstance().markAppCrash()
+            
+            var message = exception.description
+            message += "\n"
+            message += exception.callStackSymbols.description
+            Logger.error(message, file: "UncaughtExcepetion", function: "onUncaughtExcepetion", line: 0)
```
wbaumann: Will this still work for Obj-C crash stacks? I assume the NSString -> String conversions will still work

SmartReceipts/Services/PushNotificationService.swift
```diff
@@ -0,0 +1,71 @@
...
+//  Copyright © 2017 Will Baumann. All rights reserved.
+//
+
+import Foundation
+import UserNotifications
+import RxSwift
+import FirebaseMessaging
+import SwiftyJSON
+
+class PushNotificationService: NSObject {
```
wbaumann: Is it easy to add tests for this class too?

SmartReceipts/AppDelegate.swift
EvsenevDev: @wbaumann yes

SmartReceipts/Services/PushNotificationService.swift
EvsenevDev: @wbaumann for test it i need handle Push from FCM. Do you have idea how test it? I have one:
Call Delegate method with my UserData if it possible and try to subscribe to observable. What do you think?

wbaumann: @EvsenevDev Yea - I think that'd be good. Fake being logged in and verify that it calls your mock service on subscribe

PR #159, opened 2017-10-25
OCR Scans Purchase Tracking

EvsenevDev:
- Added LocalScansTracker
- [Unit Tests] Added tests for LocalScansTracker
- Added ScansPurchaseTracker
- [Unit Tests] Tests for ScansPurchaseTracker

### DIFF
--- a/SmartReceipts/Services/PurchaseService.swift
+++ b/SmartReceipts/Services/PurchaseService.swift
@@ -70,22 +70,31 @@ class PurchaseService {
 extension PurchaseService {
     func sendConfirm(transaction: SKPaymentTransaction) {
         if AuthService.shared.isLoggedIn {
-            guard let receiptURL = Bundle.main.appStoreReceiptURL else { return }
-            guard let receipt = try? Data(contentsOf: receiptURL) else { return }
-            let receiptString = receipt.base64EncodedString()
-            Logger.debug("RECEIPT: \(receiptString)")
-            let params = ["encoded_receipt": receiptString,
-                          "pay_service" : "Apple Store",
-                          "goal" : "Recognition"]
-            
-            APIAdapter.jsonBody(.post, endpoint("mobile_app_purchases"), parameters: params)
-                .subscribe(onNext: { response in
-                    let jsonRespose = JSON(response)
-                    Logger.debug(jsonRespose.description)
-                }, onError: { error in
-                    Logger.error(error.localizedDescription)
-                }).disposed(by: bag)
+            sendReceipt()
         }
+    }
+    
+    func sendReceipt() {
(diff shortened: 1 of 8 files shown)


SmartReceipts/Services/PurchaseService.swift
```diff
@@ -70,23 +70,32 @@ class PurchaseService {
...
+        let receiptString = receipt.base64EncodedString()
+        Logger.debug("RECEIPT: \(receiptString)")
+        let params = ["encoded_receipt": receiptString,
+                      "pay_service" : "Apple Store",
+                      "goal" : "Recognition"]
+        
+        APIAdapter.jsonBody(.post, endpoint("mobile_app_purchases"), parameters: params)
+            .retry(3)
+            .do(onNext: { _ in
+                ScansPurchaseTracker.shared.fetchAndPersistAvailableRecognitions()
```
wbaumann: Rather than mapping this through a `do(onNext: {}` block, we should pass it through a `flatMap`. 

As a general rule, `doOnNext` should only be used for simple side effects (eg logging, metrics, etc), while `flatMap` should be used any time you have another `Observable`. Some odd side effects can be introduced by having an inner `.subscribe().disposed(...)` block. Does that make sense?

SmartReceipts/Services/ScansPurchaseTracker.swift
```diff
@@ -0,0 +1,49 @@
...
+    
+    func initialize() {
+        AuthService.shared.loggedInObservable
+            .filter({ $0 })
+            .flatMap({ _ -> Observable<Int> in
+                return self.fetchAndPersistAvailableRecognitions()
+            }).subscribe()
+            .disposed(by: bag)
+        
+        purchaseService.sendReceipt()
```
wbaumann: Any idea if there's a way to locally mark a receipt as `uploaded`? I can open another ticket, but I was thinking (after our earlier discussion) that it may be helpful to not overload our services (even though it won't allow duplicates). Thoughts?

EvsenevDev: @wbaumann first idea it's save flag in **UserDefaults**, where 
key = receipt_[receipt] 
value = true
if UserDefaults not contains this key than we need to upload receipt.

SmartReceipts/Services/PurchaseService.swift
EvsenevDev: done

SmartReceipts/Services/ScansPurchaseTracker.swift
wbaumann: @EvsenevDev Yep - that's what I did on Android, so I think it's a good approach that's also simple

PR #160, opened 2017-10-28
OCR Configuration Module

EvsenevDev:
- Added OCR Configuration Module
- Added tests for OCR Configuration Module
- Changes for Pull Request#159, using flatMap in sendReceipt
- Added OCR Purchase flow

### DIFF
--- a/SmartReceipts/Modules/AppModules.swift
+++ b/SmartReceipts/Modules/AppModules.swift
@@ -25,4 +25,5 @@ enum AppModules: String, ViperitModule {
     case paymentMethods
     case categories
     case debug
+    case oCRConfiguration
 }
--- a/SmartReceipts/Modules/OCR Configuration/OCRConfigurationPresenter.swift
+++ b/SmartReceipts/Modules/OCR Configuration/OCRConfigurationPresenter.swift
@@ -8,8 +8,27 @@
 
 import Foundation
 import Viperit
+import RxSwift
+import StoreKit
 
 class OCRConfigurationPresenter: Presenter {
+    private let bag = DisposeBag()
+    
+    override func viewHasLoaded() {
+        interactor.requestProducts()
+            .subscribe(onNext: { [weak self] product in
+                guard let sSelf = self else { return }
+                let bag = sSelf.bag
+                let price = Observable<String>.just(product.localizedPrice)
+                product.productIdentifier == PRODUCT_OCR_10 ?
+                    price.bind(to: sSelf.view.OCR10Price).disposed(by: bag) : ()
+                
+                product.productIdentifier == PRODUCT_OCR_50 ?
+                    price.bind(to: sSelf.view.OCR50Price).disposed(by: bag) : ()
+            }).disposed(by: bag)
+        
+        
+    }
 }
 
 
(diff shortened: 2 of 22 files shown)


SmartReceipts/Modules/AppModules.swift
```diff
@@ -25,4 +25,5 @@ enum AppModules: String, ViperitModule {
     case paymentMethods
     case categories
     case debug
+    case oCRConfiguration
```
wbaumann: Wanna just name this `ocrConfiguration` as well. Or leave `OCR` in all caps? It looks kind of weird otherwise

SmartReceipts/Modules/OCR Configuration/OCRConfigurationPresenter.swift
```diff
@@ -0,0 +1,54 @@
...
+    private let bag = DisposeBag()
+    
+    override func viewHasLoaded() {
+        interactor.requestProducts()
+            .subscribe(onNext: { [weak self] product in
+                guard let sSelf = self else { return }
+                let bag = sSelf.bag
+                let price = Observable<String>.just(product.localizedPrice)
+                product.productIdentifier == PRODUCT_OCR_10 ?
+                    price.bind(to: sSelf.view.OCR10Price).disposed(by: bag) : ()
```
wbaumann: Let's try to avoid this inner `disposedBy` block. What's the purpose of `price.bindTo`? Can `flatMap` to this second Observable instead:

```
interactor.requestProducts()
   .flatMap() ({product in
        if (product.identifer == OCR_10) {
            return price.bind(to: view.OCR10)
        } ...
    }
    .subscribe() {
        // Log or something
    }
```

Though I guess it depends what `price.bind` does. This just sets the text right?

SmartReceipts/Views/ScansPurchaseButton.swift
```diff
@@ -0,0 +1,50 @@
...
+    @IBOutlet fileprivate weak var activityIndicator: UIActivityIndicatorView!
+ 
+    private let bag = DisposeBag()
+    
+    override func awakeFromNib() {
+        layer.cornerRadius = 7
+        
+        rx.controlEvent(.touchDown)
+            .subscribe(onNext: { [unowned self] in
+                self.backgroundColor  = #colorLiteral(red: 0.3334693832, green: 0.08451540429, blue: 0.4416083583, alpha: 1)
```
wbaumann: Let's move these into the `AppTheme` (or an extension). Maybe name it like `ButtonStyle1_Pressed` vs `ButtonStyle1_UnPressed`

SmartReceipts/Modules/AppModules.swift
EvsenevDev: I've tried with name **OCRConfiguration**, it works. Let's use it.
I called it **oCRConfiguration** because Viperit have some requirements for names.
For example: ocrConfiguration not working.

SmartReceipts/Modules/OCR Configuration/OCRConfigurationPresenter.swift
EvsenevDev: Right

SmartReceipts/Views/ScansPurchaseButton.swift
EvsenevDev: done

PR #161, opened 2017-10-29
Dropdown Menu

EvsenevDev:
* Added Dropdown Menu
* OCR Configuration integrated with Auth
* Changes for PR#160
* Deleted **documentsPath** logs


SmartReceipts/Modules/Auth/AuthInteractor.swift
```diff
@@ -33,7 +33,10 @@ class AuthInteractor: Interactor {
                         }
                         return Observable.never()
                     }).filter({ $0 != ""})
-                    .bind(to: self.presenter.successLogin)
+                    .map({ [weak self] in
+                        self?.presenter.successAuthSubject.onNext()
```
wbaumann: For these, since you're just returning the initial value (ie `return $0`), you can use a `do(onNext{` instead

SmartReceipts/Modules/Trips/TripsPresenter.swift
```diff
@@ -16,31 +16,54 @@ class TripsPresenter: Presenter {
...
-        tripDetailsSubject.subscribe(onNext: { trip in
-            self.router.openDetails(trip: trip)
-        }).disposed(by: disposeBag)
+        view.autoScansTap
+            .filter({ !AuthService.shared.isLoggedIn })
+            .subscribe(onNext: {
+                let authModule = self.router.openAuth()
+                _ = authModule.successAuth
+                    .map({ authModule.close() })
+                    .delay(0.6, scheduler: MainScheduler.instance)
```
wbaumann: Just curious, why do we introduce a delay here?

SmartReceipts/Modules/Auth/AuthInteractor.swift
EvsenevDev: @wbaumann but you said logic in **do()** isn't good way.

SmartReceipts/Modules/Trips/TripsPresenter.swift
EvsenevDev: @wbaumann we need to wait 2 animations:
1) Hiding PendingHUD
2) Closing AuthView
We can't open OCR while Auth is closing.

SmartReceipts/Modules/Auth/AuthInteractor.swift
wbaumann: @EvsenevDev - Apologies for the confusion :)

`do` is good for side-effects (eg logging, analytics, etc). This is actually a good case for it on second thought, since this is effectively a side effect, since `onNext` happens as a side effect of the success of this first stream

SmartReceipts/Modules/Trips/TripsPresenter.swift
wbaumann: @EvsenevDev Is it possible to trigger once the Hud and AuthView have closed? Or do we not get a message when they do?

EvsenevDev: @wbaumann no, we can't we not have a message and we need show OCR on dealloc AuthView, for clear view stack
Standart animation duration = 0.3 (2 animations = 0.6)

SmartReceipts/Modules/Auth/AuthInteractor.swift
EvsenevDev: @wbaumann ok, thanks. I will change it in next PR

SmartReceipts/Modules/Trips/TripsPresenter.swift
wbaumann: @EvsenevDev Sounds good. Thanks!

PR #169, opened 2017-12-03
PDF Updates

EvsenevDev:
- Fixed Min. Receipt Price field in Settings
- Add **PDFFontStyle** enum to AppTheme
- Changed PDF Fonts
- PDF Part re-written to Swift

PR #170, opened 2017-12-09
PDF Report Upgrade

EvsenevDev:
- Generators migrated to Swift
- Updated Report Header
- Removed currency symbol in Report
- Fixed PDF columns squeeze
- Added missing comment from PR #169 
- Fixed Open Settings by Show Upgrade Options button in Excahge Rate Cell
- Fixed Crash on TableView delete item


SmartReceipts/Reports/Generators/TripCSVGenerator.swift
```diff
@@ -0,0 +1,51 @@
...
+            Logger.error("CSV write error: \(error.localizedDescription)")
+           return false
+        }
+        return true
+    }
+    
+    private func generateContent() -> String {
+        let content = NSMutableString(string: String.byteOrderMarked)
+        appendReceiptsTable(content)
+        appendReceiptsTable(content)
```
wbaumann: This should be `appendDistanceTable`, correct?

SmartReceipts/Reports/Generators/TripFullPDFGenerator.swift
```diff
@@ -0,0 +1,119 @@
...
+        appenSummaryAndTables()
+
+        // always render fullPageElements and images in protrait orientation
+        pdfRender.landscapePreferred = false
+        pdfRender.startNextPage()
+        appendImages()
+        return pdfRender.renderPages()
+    }
+    
+    private func appenSummaryAndTables() {
```
wbaumann: Typo in name. It should be `appendSummaryAndTables` instead of `appenSummaryAndTables` (missing a `d`).

SmartReceipts/Reports/Generators/TripFullPDFGenerator.swift
```diff
@@ -0,0 +1,119 @@
...
+        var recs = receipts()
+        
+        for receipt in recs {
+            if reportOnlyReimbursable && !receipt.isReimbursable {
+                continue
+            }
+            grossTotal.addPrice(receipt.targetPrice())
+            receiptTotal.addPrice(receipt.targetPrice())
+            noTaxesTotal.addPrice(receipt.targetPrice())
+            noTaxesTotal.subtractPrice(receipt.targetTax())
```
wbaumann: For each of these, I think we need to adjust the logic for Pre-Tax Price vs Post-Tax Price like here:

https://github.com/wbaumann/SmartReceiptsLibrary/blob/master/app/src/main/java/co/smartreceipts/android/workers/reports/pdf/pdfbox/ReceiptsTotals.java#L56

This setting is pretty confusing, and I need to do a better idea of clarifying what it does. The rough idea is that the user only has access to two fields when adding a new receipt:

* `Price`
* `Tax`

The `Tax` field will always contain the tax value, but the `Price` value could either be the Total (ie with Tax) or the SubTotal (ie without Tax). If the user has set the `Assume Price Is Pre Tax` setting, it means that she is using the subtotal. If it is not set (ie false), it assumes she is using the total.

For a more concrete example, let's assume the subtotal is `$5`, Tax is `$1`, and total is `$6`. If the user inputs the price as `$5` and the tax as `$1`, the user is on the pre-tax setting. If the user enters the price as `$6` and the tax as `$1`, the user is on the post-tax setting (ie pre-tax is false). As a result, we need to adjust these various totals based on this setting

PR #171, opened 2017-12-11
Changes for PR#170

EvsenevDev:
- Changes for PR#170
- [Fix] CSV Generator

P.S. Now PDF looks like on Android


SmartReceipts/Extensions/String+Extensions.swift
```diff
@@ -19,10 +19,6 @@ extension String {
         return URL(fileURLWithPath: self)
     }
     
-    static var byteOrderMarked: String {
-        return "\\357\\273\\277"
```
wbaumann: How come we got rid of this one?

PR #172, opened 2017-12-17
Layout Update

EvsenevDev:
- Fixed Receipts NavBar Title layout
- Fixed Plus button is never hidden by our ad
- Fixed Menu Button layout
- Fixed FetchedPlaceholder rotation layout
- Fixed Menu disappear after navigation

### DIFF
--- a/SmartReceipts/ViewControllers/Abstract/Customization/AppTheme.swift
+++ b/SmartReceipts/ViewControllers/Abstract/Customization/AppTheme.swift
@@ -10,6 +10,8 @@ import Foundation
 
 class AppTheme: NSObject {
     
+    static let appTitle = "Smart Receipts"
+    
     // Colors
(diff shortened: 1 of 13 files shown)


SmartReceipts/ViewControllers/Abstract/Customization/AppTheme.swift
```diff
@@ -10,6 +10,8 @@ import Foundation
 
 class AppTheme: NSObject {
     
+    static let appTitle = "Smart Receipts"
```
wbaumann: What's this for?

EvsenevDev: @wbaumann it's for main navigation bar title. I don't like to hardcode it in ViewController.

wbaumann: @EvsenevDev Makes sense. Could you make it a localizable string instead? In case we ever change it

PR #173, opened 2017-12-24
Report Table Footers

EvsenevDev:
- Added report table footer for receipts
- Added report table footer for distances

### DIFF
--- a/SmartReceipts/Model/Receipt/ReceiptColumnNetExchangedPricePlusTax.swift
+++ b/SmartReceipts/Model/Receipt/ReceiptColumnNetExchangedPricePlusTax.swift
@@ -16,4 +16,13 @@ class ReceiptColumnNetExchangedPricePlusTax: ReceiptColumn {
         
         return receipt.exchangedNetPriceAsString()
     }
+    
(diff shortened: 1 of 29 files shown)


SmartReceipts/Model/Receipt/ReceiptColumnNetExchangedPricePlusTax.swift
```diff
@@ -16,4 +16,13 @@ class ReceiptColumnNetExchangedPricePlusTax: ReceiptColumn {
         
         return receipt.exchangedNetPriceAsString()
     }
+    
+    override func value(forFooter rows: [Any]!, forCSV: Bool) -> String! {
+        var total = NSDecimalNumber.zero
+        for rec in rows as! [WBReceipt] {
+            total = total.adding(rec.exchangedPrice()!.amount)
+            total = total.adding(rec.exchangedTax()!.amount)
```
wbaumann: For this one, you'll also need to update it to check price vs pre-tax price:

https://github.com/wbaumann/SmartReceiptsLibrary/blob/master/app/src/main/java/co/smartreceipts/android/model/impl/columns/receipts/ReceiptNetExchangedPricePlusTaxColumn.java#L31

PR #175, opened 2017-12-31
Added Migration Service and migration Illigal Trip Names

EvsenevDev:


SmartReceipts/Services/MigrationService.swift
```diff
@@ -0,0 +1,39 @@
...
+import FMDB
+
+fileprivate let migrationKey = "migration_version_key"
+
+class MigrationService {
+    
+    func migrate() {
+        
+        //MARK: Migration 0 -> 1
+        migrate(from: 0, to: 1, {
```
wbaumann: Would it be possible to add tests for this one too?

PR #176, opened 2018-01-07
Image Upload optimization

EvsenevDev:
@wbaumann 
I've added optimization to Image Upload, but this works for local saving too. It's ok? Or need changes just for uploading?


wbaumann:
@EvsenevDev Nope - works great for both local and remote. Thanks!

PR #177, opened 2018-01-16
Upgrade reports

EvsenevDev:
**Completed Tasks:**
- Add tests for Migration Service
- Add custom Eureka Cell - DescribedSwitchCell
- Add new Plus settings (separateByCategories, categoricalSummation, omitDefaultPdfTable)
- Add the a "Plus" setting to "Include Categorical Summation In Reports" and add the corresponding behavior to our CSV and Full Page Pdf reports
- Add the a "Plus" setting to "Separate Payments by Category" and add the corresponding behavior to our CSV and Full Page Pdf reports
- Add the a "Plus" setting to "Omit Default PDF Table" and add the corresponding behavior to our Full Page Pdf reports


SmartReceipts/Modules/Settings/SettingsFormView.swift
```diff
@@ -105,6 +106,50 @@ class SettingsFormView: FormViewController {
                                         message: LocalizedString("settings.pdf.footer.pro.message.body")))
             }
         })
+        
+        <<< DescribedSwitchRow() { row in
+            row.title = LocalizedString("settings.pro.payments.by.category.title")
+            row.value = WBPreferences.separatePaymantsByCategory()
+            row.subtitle = row.value! ?
+                LocalizedString("settings.pro.payments.by.category.description.on") :
+                LocalizedString("settings.pro.payments.by.category.description.off")
```
wbaumann: Is it possible to set description_on vs description_off as view properties? That way we never manually set it, but it triggers based off whenever the row.value field changes?

SmartReceipts/Modules/Settings/SettingsFormView.swift
```diff
@@ -105,6 +106,50 @@ class SettingsFormView: FormViewController {
...
+        }.onChange({ row in
+            WBPreferences.setSeparatePaymantsByCategory(row.value!)
+            row.subtitle = row.value! ?
+                LocalizedString("settings.pro.payments.by.category.description.on") :
```
wbaumann: If we do it this way, we don't need to support all the `onChange` methods and can instead just apply these to the view when we first initialize it

SmartReceipts/PDF/PDFReportTable.swift
```diff
@@ -73,11 +85,16 @@ class PDFReportTable: UIView {
...
         columnsWidth = divideRemainingWidth(columns: columnsWidth, titleWidth: titleWidth)
         
-        var yOffset: CGFloat = 0
+        var yOffset: CGFloat = 10
         // Don't add header on view where already added
         // Happens when we decide that not enough rows in page and push all to next page.
         // In that case original table is reused, just additional rows added
         
+        if let tableTitle = title {
+            yOffset = setTable(title: tableTitle, yOffset: yOffset) + 8
```
wbaumann: Can we make the `+ 8` into a const? something like `tableSeparatorSize`

SmartReceipts/PDF/PDFReportTable.swift
```diff
@@ -73,11 +85,16 @@ class PDFReportTable: UIView {
         
         columnsWidth = divideRemainingWidth(columns: columnsWidth, titleWidth: titleWidth)
         
-        var yOffset: CGFloat = 0
+        var yOffset: CGFloat = 10
```
wbaumann: Can we make the `10` here a named value. So

`yOffset: CGFloat = SOME_CONSTANT`

SmartReceiptsTests/Reports/Generators/TripCSVGeneratorTest.swift
```diff
@@ -21,18 +21,18 @@ class TripCSVGeneratorTest: SmartReceiptsTestsBase {
     }
```
wbaumann: Can we add tests to confirm the category grouping stuff works? It'd be great to confirm that generating those tables is reflected here properly

SmartReceipts/Modules/Settings/SettingsFormView.swift
EvsenevDev: I'm thought about it too ;)

EvsenevDev:
@wbaumann all done

PR #178, opened 2018-01-19
Updated Ads

EvsenevDev:
- Optimized Purchase Service
- Removed Google Pods and changed to Firebase
- Updated AdMob
- Removed GoogleAnalytics
- Added UpsellBannerView and Upsell
- Updated Ad Flow


SmartReceipts/Services/PurchaseService.swift
```diff
@@ -18,6 +18,13 @@ let PRODUCT_PLUS = "ios_plus_sku_2"
 class PurchaseService {
     private var plusSubsribtionProduct: SKProduct?
     fileprivate let bag = DisposeBag()
+    static fileprivate var cachedProducts = [SKProduct]()
+    
+    func cacheProducts() {
```
wbaumann: Does this ever get updated? For instance, what happens if we make a new purchase?

EvsenevDev: @wbaumann it's just `Product` it's not Transaction or Purchase

PR #179, opened 2018-01-27
Import Updates

EvsenevDev:
- Added Import From Files feature
- Changed the name of the iOS photo item from "Image" to "Picture"
- "Add Receipt" actions order changed
- Added PDF Support to Scan flow
- Added Cancel button to Import actions
- Added Cancel button to FilesViewController
- Some Appearance changes for FilesViewController


SmartReceipts/Model/Scan.swift
```diff
@@ -12,19 +12,37 @@ import SwiftyJSON
...
+        parseJSON(json)
+    }
+    
+    convenience init(json: JSON, image: UIImage) {
+        self.init(image: image)
+        parseJSON(json)
+    }
+    
+    private func cache(image: UIImage) -> URL {
+        let imgData = UIImageJPEGRepresentation(image, 0.85)!
```
wbaumann: Can we move the `0.85` to a constant value (eg `imageCompressionRatio`)? Then we can share it between this and the other classes that use it (there's one more, right?)

SmartReceipts/Modules/Receipts/ReceiptsRouter.swift
```diff
@@ -71,7 +71,23 @@ class ReceiptsRouter: Router {
             }).subscribe(onNext: { [unowned self] scan in
                 hud?.hide()
                 self.openEditModule(with: scan)
-            }).disposed(by: bag)
+            })
+    }
+    
+    func openImportReceiptFile() {
+        var hud: PendingHUDView?
+        if #available(iOS 11.0, *) {
```
wbaumann: Does this also prevent the button from appearing altogether on pre iOS 11

SmartReceipts/Services/ScanService.swift
```diff
@@ -47,59 +47,94 @@ class ScanService {
...
-                        .do(onNext: { [weak self] _ in
-                            self?.statusSubject.onNext(.completed)
-                            AnalyticsManager.sharedManager.record(event: Event.ocrRequestSucceeded())
-                            ScansPurchaseTracker.shared.decrementRemainingScans()
-                        })
-                }).catchError({ error -> Observable<Scan> in
-                    return Observable<Scan>.just(Scan(image: UIImage()))
-                })
+        
+            return nil
```
wbaumann: Why `return nil` here? Was this just for testing

wbaumann: Should this be return `scanFrom`?

SmartReceipts/Supporting Files/pl.lproj/Localizable.strings
```diff
@@ -288,8 +288,14 @@
 "receipt.column.report.price.minus.tax" = "Cena Minus Tax";
 "receipt.column.report.price.exchanged.minus.tax" = "Cena Minus Tax (Exchanged)";
 "receipt.column.user.id" = "ID Użytkownika";
-"receipt.floaty.item.text.only" = "Text-Only";
-"receipt.floaty.item.image" = "Image";
+"receipt_floaty_item_text_only" = "Text-Only";
```
wbaumann: Thanks :)

SmartReceipts/Utils/Import/ReceiptFilePicker.swift
```diff
@@ -0,0 +1,131 @@
...
+//  Copyright © 2018 Will Baumann. All rights reserved.
+//
+
+import RxSwift
+
+fileprivate let JPEG_TYPE = "public.jpeg"
+fileprivate let PNG_TYPE = "public.png"
+fileprivate let PDF_TYPE = "com.adobe.pdf"
+
+class ReceiptFilePicker: NSObject {
```
wbaumann: Can we add UTs here? It'd be nice to test the iOS 11 vs 10 flows

EvsenevDev: @wbaumann If i right understand you, I think we can't do something good for this, because it's just wrapper for UI.

SmartReceipts/Modules/Receipts/ReceiptsRouter.swift
EvsenevDev: @wbaumann, compiler needs this mark each time if we use code for iOS 11 and our deployment target is 10

SmartReceipts/Model/Scan.swift
EvsenevDev: @wbaumann yep, sorry :)

SmartReceipts/Modules/Receipts/ReceiptsRouter.swift
EvsenevDev: @wbaumann but here is my mistake, yep, you are right, thanks. This class was just for iOS 11, and then i changed this class to universal, buе forgot to remove this mark. Thanks.

SmartReceipts/Services/ScanService.swift
EvsenevDev: This means if return nil chain is interrupted and we not start scanning process we return Observable.just with empty scan. nil it's about fail. It's bad approach with Rx?

wbaumann: Ah okay. Can we use either `Observable.empty()` or `Observable.error(...)` for this case?

SmartReceipts/Utils/Import/ReceiptFilePicker.swift
wbaumann: Okay makes sense. Thanks :)

SmartReceipts/Services/ScanService.swift
EvsenevDev: @wbaumann i did remove this function, it's not needed i wrote it before adding scan with ReceiptDocument

wbaumann:
@EvsenevDev Thanks :)

PR #182, opened 2018-02-01
Menu & SharedLocalizable

EvsenevDev:
- Menu Button now opening the Action Sheet
- Added **SharedLocalizable**


SmartReceipts/Modules/Trips/TripsView.swift
```diff
@@ -59,7 +59,7 @@ final class TripsView: FetchedTableViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         title = AppTheme.appTitle
-        configureMenu()
+        //configureMenu()
```
wbaumann: Can you add a comment above why this was commented out? With a link to the open issue on that GitHub repo that you found?

EvsenevDev: @wbaumann ok

PR #189, opened 2018-02-18
[Fix] Added safe checking WBTrip fields in PDF Generator

EvsenevDev:


SmartReceipts/Reports/Generators/TripFullPDFGenerator.swift
```diff
@@ -82,11 +82,11 @@ class TripFullPDFGenerator: TripImagesPDFGenerator {
         let endDate = dateFormatter.formattedDate(trip.endDate, in: trip.endTimeZone)!
         pdfRender.appendHeader(row: String(format: LocalizedString("pdf.report.from.to.label.base"), startDate, endDate))
         
-        if WBPreferences.trackCostCenter() && !trip.costCenter.isEmpty {
+        if let costCenter = trip.costCenter, WBPreferences.trackCostCenter() && !trip.costCenter.isEmpty {
```
wbaumann: @EvsenevDev I'm going to merge in now. Thanks! Could you also add unit tests for when `costCenter` and `comment` are `nil`? That way we'll avoid these crashes in the future if we ever make changes here. Thanks!

EvsenevDev: @wbaumann `nil` is ok for `comment` and `costCenter`. It's just Objective-C to Swift factor. I think it's not need Unit Test. I'm not imagine Unit Test code for nil checking :) But if we want to always return empty string instead nil. We can add unit test for it. But i think `nil` is ok for empty fields in WBTrip

wbaumann: @EvsenevDev I agree that `nil` is okay for both, since it's from Objective-C

I meant a test for `TripFullPDFGenerator.swift` to confirm that it works (ie doesn't crash) when `comment` or `costCenter` is `nil`

EvsenevDev: @wbaumann ok, i'll add it after Inline Picker with Button

wbaumann:
Sounds good. Thank you :)

PR #197, opened 2018-03-11
Added Tax Picker to Edit Receipt Module

EvsenevDev:


SmartReceipts/SmartReceipts-Info.plist
```diff
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>${PRODUCT_NAME}</string>
```
wbaumann: Can we change the `${PRODUCT_NAME}` instead? Or is this too much work?

wbaumann: Synced offline. We can, but then we cannot localize this later.

PR #198, opened 2018-03-14
Some changes and fixes

EvsenevDev:
- Allow CSV reports to only include distance entries
- Added opening Settings on Distance section for try to generate empty CSV
- [Fix] Update Settings after Restore Subscription
- Added metrics to Generate (nothing to generate for CSV)
- Open app on last selected Trip
- **Crashlytics migration**

### DIFF
--- a/SmartReceipts/Modules/Debug/DebugFormView.swift
+++ b/SmartReceipts/Modules/Debug/DebugFormView.swift
@@ -11,6 +11,7 @@ import Eureka
 import RxSwift
 import RxCocoa
 import Toaster
+import Crashlytics
 
 fileprivate let SCAN_ROW = "ScanRow"
 
@@ -86,6 +87,21 @@ class DebugFormView: FormViewController {
             
         <<< ScanRow(SCAN_ROW)
         
+        +++ Section("Crashlytics")
+        <<< SwitchRow() { row in
+            row.title = "Crashlytics 'DEBUG MODE'"
+            row.value = Crashlytics.sharedInstance().debugMode
+        }.onChange({ row in
+            guard let value = row.value else { return }
+            Crashlytics.sharedInstance().debugMode = value
+        })
+            
+        <<< ButtonRow() { row in
+            row.title = "Initiate crash"
+        }.onCellSelection({ _, _ in
+            Crashlytics.sharedInstance().debugMode = false
+            Crashlytics.sharedInstance().crash()
+        })
     }
     
     var scanObserver: AnyObserver<Scan> {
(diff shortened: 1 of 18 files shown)


SmartReceipts/Modules/Debug/DebugFormView.swift
```diff
@@ -86,6 +87,21 @@ class DebugFormView: FormViewController {
...
+        }.onChange({ row in
+            guard let value = row.value else { return }
+            Crashlytics.sharedInstance().debugMode = value
+        })
+            
+        <<< ButtonRow() { row in
+            row.title = "Initiate crash"
+        }.onCellSelection({ _, _ in
+            Crashlytics.sharedInstance().debugMode = false
+            Crashlytics.sharedInstance().crash()
```
wbaumann: Nice :)

I forwarded you the crash email

SmartReceipts/Modules/Trips/TripsInteractor.swift
```diff
@@ -24,6 +26,18 @@ class TripsInteractor: Interactor {
...
         return Database.sharedInstance().createUpdatingAdapterForAllTrips()
     }
+    
+    func markLastOpened(trip: WBTrip) {
+        UserDefaults.standard.set(trip.name, forKey: LAST_OPENED_TRIP_KEY)
+    }
+    
+    var lastOpenedTrip: WBTrip? {
+        guard let tripName = UserDefaults.standard.value(forKey: LAST_OPENED_TRIP_KEY) as? String else { return nil }
+        if let trip = Database.sharedInstance().tripWithName(tripName) {
```
wbaumann: Given the database call, should we wrap this in an `Maybe` so we can fetch from the background?

SmartReceipts/Modules/Trips/TripsPresenter.swift
```diff
@@ -23,6 +23,10 @@ class TripsPresenter: Presenter {
         interactor.configureSubscribers()
         executeFor(iPhone: {}, iPad: { router.openNoTrips() })
         
+        if !Database.sharedInstance().allTrips().isEmpty, let lastTrip = interactor.lastOpenedTrip {
```
wbaumann: In conjunction with this piece - I don't want to free the UIThread with a disk read if possible (note: Android may also do this wrong)

wbaumann: Also - I don't think we `!Database.sharedInstance().allTrips().isEmpty`, since `lastTrip` will be `nil` if we don't find it in the database

PR #199, opened 2018-03-17
Changes 1.10

EvsenevDev:
- PR #198 Changes
- [Fix] Increase the PDF header text quality
- [Fix] Bug in which file names do not change with the receipt names
- Unwind the "Cost Center" optional

### DIFF
--- a/SmartReceipts/Modules/Trips/TripsPresenter.swift
+++ b/SmartReceipts/Modules/Trips/TripsPresenter.swift
@@ -23,9 +23,11 @@ class TripsPresenter: Presenter {
         interactor.configureSubscribers()
         executeFor(iPhone: {}, iPad: { router.openNoTrips() })
         
-        if !Database.sharedInstance().allTrips().isEmpty, let lastTrip = interactor.lastOpenedTrip {
-            router.openDetails(trip: lastTrip)
-        }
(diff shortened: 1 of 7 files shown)


SmartReceipts/Modules/Trips/TripsPresenter.swift
```diff
@@ -23,9 +23,11 @@ class TripsPresenter: Presenter {
         interactor.configureSubscribers()
         executeFor(iPhone: {}, iPad: { router.openNoTrips() })
         
-        if !Database.sharedInstance().allTrips().isEmpty, let lastTrip = interactor.lastOpenedTrip {
-            router.openDetails(trip: lastTrip)
-        }
+        interactor.lastOpenedTrip
```
wbaumann: This only gets called the first time you launch this, correct? For instance, you couldn't have the following:

1. Launch `TripsController`
2. Auto-open receipts for this trip
3. Click back
4. Auto-open again

EvsenevDev: @wbaumann Hi,
lastOpenedTrip always create new Maybe, it's not stored property or subject, we haven't issue with second reopen this, if you about it.

PR #202, opened 2018-04-01
Added custom sorting for receipts, categories, and columns

EvsenevDev:
- PR #198 Changes
- [Fix] Increase the PDF header text quality
- [Fix] Bug in which file names do not change with the receipt names
- Unwind the "Cost Center" optional
- Added custom_order_id field to receipts table
- Added v16 Migration
- Changed Receipts sort to CUSTOM_ORDER_ID and updated Swap
- Added support for category sorting via a custom order id
- Changed the categories to have a unique number as primary key
- Changed approach with Categories Primary Key migration
- Added support for column sorting via a custom order id


SmartReceipts/AppDelegate.swift
```diff
@@ -29,7 +29,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     func applicationDidFinishLaunching(_ application: UIApplication) {
         AppDelegate.instance = self
-        FirebaseApp.configure()
```
wbaumann: How come we removed this line? Or no longer required?

SmartReceipts/Modules/Columns/ColumnsInteractor.swift
```diff
@@ -22,10 +22,25 @@ class ColumnsInteractor: Interactor {
         return Observable<[Column]>.just(columns as! [Column])
     }
     
-    func update(columns: [Column], forCSV: Bool) {
+    func addColumn(_ column: Column, isCSV: Bool) {
         let db = Database.sharedInstance()!
-        let result = forCSV ? db.replaceAllCSVColumns(with: columns) : db.replaceAllPDFColumns(with: columns)
-        Logger.info("Saving columns. Result: \(result)")
+        let orderId = isCSV ? db.nextCustomOrderIdForCSVColumn() : db.nextCustomOrderIdForPDFColumn()
```
wbaumann: One note here, we default this to `Long.MAX_VALUE` on Android, but I think your approach is a bit safer.

SmartReceipts/Modules/Edit Receipt/EditReceiptInteractor.swift
```diff
@@ -77,6 +78,22 @@ class EditReceiptInteractor: Interactor {
         }
     }
     
+    private func replaceImageIfNeeded(receipt: WBReceipt) {
+        if receipt.hasPDF() || receipt.hasImage() {
+            let currentPath = receipt.imageFilePath(for: receipt.trip)
+            let tripFolder = currentPath.asNSString.deletingLastPathComponent
+            let newFileName = String(format: "%tu_%@.%@", receipt.objectId, receipt.omittedName, currentPath.asNSString.pathExtension)
+            let newPath = "\(tripFolder)/\(newFileName)"
+            receipt.setImageFileName(newFileName)
```
wbaumann: We should probably eventually change this method signature to be `setFileName` instead of `setImageFileName`

SmartReceipts/Persistence/Helpers/Database+Categories.m
```diff
@@ -60,14 +63,101 @@ - (BOOL)updateCategory:(WBCategory *)category {
...
+
+- (BOOL)swapCategory:(WBCategory *)categoryOne withCategory:(WBCategory *)categoryTwo usingDatabase:(FMDatabase *)database {
+    return [self setCustomOrderId:categoryOne.customOrderId forCategory:categoryTwo usingDatabase:database] &&
+           [self setCustomOrderId:categoryTwo.customOrderId forCategory:categoryOne usingDatabase:database];
+}
+
+- (BOOL)setCustomOrderId:(NSInteger)customOrderId forCategory:(WBCategory *)category usingDatabase:(FMDatabase *)database {
+    DatabaseQueryBuilder *update = [DatabaseQueryBuilder updateStatementForTable:CategoriesTable.TABLE_NAME];
+    [update addParam:CategoriesTable.COLUMN_CUSTOM_ORDER_ID value:@(customOrderId)];
+    [update where:CategoriesTable.COLUMN_NAME value:category.name];
```
wbaumann: Let's change this to use `where:CategoriesTable.COLUMN_ID` instead, since that's our new primary key

SmartReceipts/Persistence/Helpers/Database+Categories.m
```diff
@@ -60,14 +63,101 @@ - (BOOL)updateCategory:(WBCategory *)category {
...
+    if (result) {
+        NSArray *changedModels = [self categoriesBetweenCategoryOne:categoryOne categoryTwo:categoryTwo];
+        [self notifySwapOfModels:changedModels];
+    }
+    
+    return result;
+}
+
+- (BOOL)swapCategory:(WBCategory *)categoryOne withCategory:(WBCategory *)categoryTwo usingDatabase:(FMDatabase *)database {
+    return [self setCustomOrderId:categoryOne.customOrderId forCategory:categoryTwo usingDatabase:database] &&
```
wbaumann: I know we don't do this anywhere else in the app, but we should probably start using transactions to mark when it started and was successful like:

https://stackoverflow.com/a/14542469/4250037

Database transactions allow us to ensure that multiple operations can occur atomically, so we can roll back if either update fails

SmartReceipts/Persistence/Helpers/Database+Categories.m
```diff
@@ -60,14 +63,101 @@ - (BOOL)updateCategory:(WBCategory *)category {
...
+                            @" WHERE ", CategoriesTable.COLUMN_CUSTOM_ORDER_ID,
+                            @" BETWEEN ", @(minId).stringValue, @" AND ", @(maxId).stringValue,
+                            @" ORDER BY ", CategoriesTable.COLUMN_CUSTOM_ORDER_ID, @" ASC"];
+    NSString *rawQuery = [components componentsJoinedByString:@""];
+    DatabaseQueryBuilder *select = [DatabaseQueryBuilder rawQuery:rawQuery];
+    FetchedModelAdapter *fetched = [self createAdapterUsingQuery:select forModel:[WBCategory class]];
+    return fetched.allObjects;
+}
+
+- (BOOL)reorderCategory:(WBCategory *)categoryOne withCategory:(WBCategory *)categoryTwo usingDatabase:(FMDatabase *)database {
```
wbaumann: How does this differ from `swapCategory` above?

SmartReceipts/Persistence/Helpers/Database+Columns.m
```diff
@@ -65,10 +67,85 @@ - (BOOL)replaceAllColumnsInTable:(NSString *)tableName columns:(NSArray *)column
...
+    
+    __block BOOL result;
+    [self.databaseQueue inDatabase:^(FMDatabase *db) {
+        result = [self reorderColumn:columnOne withColumn:columnTwo table:table usingDB:db];
+    }];
+    
+    return result;
+}
+
+- (BOOL)reorderColumn:(Column *)columnOne withColumn:(Column *)columnTwo table:(NSString *)table usingDB:(FMDatabase *)db {
```
wbaumann: Or why does this one not have `swapColumn`

SmartReceipts/Persistence/Migrations/DatabaseCreateAtVersion11.m
```diff
@@ -82,6 +82,7 @@ - (BOOL)insertDefaultCategoriesIntoDatabase:(Database *)database {
 
     for (NSUInteger i = 0; i < cats.count - 1; i += 2) {
         WBCategory *category = [[WBCategory alloc] initWithName:cats[i] code:cats[i + 1]];
+        category.customOrderId = i/2;
```
wbaumann: Why the divide by 2 here?

SmartReceipts/Persistence/Migrations/DatabaseUpgradeToVersion16.swift
```diff
@@ -0,0 +1,105 @@
...
+import Foundation
+
+class DatabaseUpgradeToVersion16: DatabaseMigration {
+    override func version() -> UInt {
+        return 16
+    }
+    
+    override func migrate(_ database: Database) -> Bool {
+        
+        return addCustomOrderIdToReceipts(database) &&
```
wbaumann: Rather than starting with `addCustomOrderIdToReceipts`, we should end with an receipts table migration script similar to `CREATE TABLE categories_backup` stuff below. We need to do this, since the receipts table references the categories table, but we've changed that category primary key from `name` to `id`. 

https://github.com/wbaumann/SmartReceiptsLibrary/blob/7a95e0ea9139c068f15302fe93d8940ec6370c20/app/src/main/java/co/smartreceipts/android/persistence/database/tables/ReceiptsTable.java#L232

All the rest looks good, and your solution of the `= ROWID` is a good one :)

SmartReceipts/AppDelegate.swift
EvsenevDev: We have this code [here](https://github.com/wbaumann/SmartReceiptsiOS/blob/master/SmartReceipts/Analytics/FirebaseAnalytics.swift#L19), but i think you're right, if we remove **FirebaseAnalytics**, Firebase will be disabled totally.

SmartReceipts/Persistence/Helpers/Database+Categories.m
EvsenevDev: Swap it's just change order for two categories.
But reorder it's like:
We have A1, B2, C3, D4, E5, and we want to move A1 to C3 then we get B1, C2, A3, D4, E5
Hope i good explained this :)

SmartReceipts/Persistence/Helpers/Database+Columns.m
EvsenevDev: We haven't swap feature for this, we have Swap just for receipts, but  by mistake i wrote this logic for Categories and decided to leave this, maybe it will be useful in future :) If you want i can remove it for Categories.

SmartReceipts/Persistence/Migrations/DatabaseCreateAtVersion11.m
EvsenevDev: Because in for loop we have **i += 2**  and i want to order like 0,1,2,3...
It's old code i just added my part :)

SmartReceipts/Persistence/Helpers/Database+Categories.m
wbaumann: @EvsenevDev Ah yes - makes complete sense :)

SmartReceipts/Persistence/Helpers/Database+Columns.m
wbaumann: Let's remove it for now for `Categories` to avoid confusion. It'll be easy to add back in the future if we need :)

SmartReceipts/Persistence/Migrations/DatabaseCreateAtVersion11.m
wbaumann: Ah okay. I have no idea why it does `i += 2` haha. Thanks :)

SmartReceipts/Model/WBReceipt.m
```diff
@@ -55,12 +55,13 @@ - (id)initWithId:(NSUInteger)rid
...
   extraEditText2:(NSString *)extraEditText2
-  extraEditText3:(NSString *)extraEditText3 {
+  extraEditText3:(NSString *)extraEditText3
+   customOrderId:(NSInteger)customOrderId {
     self = [super init];
     if (self) {
         _objectId = rid;
         _name = name;
-        _category = category;
+        _categoryId = category.objectId;
```
wbaumann: We're passing in the `category` here, right? Can we just save it instead of the `category.objectId`?

SmartReceipts/Model/WBReceipt.m
```diff
@@ -105,22 +106,26 @@ -(NSTimeZone*)timeZone {
...
     _date = date;
+    _customOrderId = date.timeIntervalSince1970;
 
     if (!self.originalDate) {
         [self setOriginalDate:date];
     }
 }
 
+- (WBCategory *)category {
+    return [Database.sharedInstance categoryByID:_categoryId];
```
wbaumann: The reason for my comment above about `category` vs `_categoryId = category.objectId;` is that we can just return the category directly instead of making a call to disk via `[Database.sharedInstance categoryByID:_categoryId];`

SmartReceipts/Model/WBReceipt.m
```diff
@@ -211,6 +215,8 @@ - (void)loadDataFromResultSet:(FMResultSet *)resultSet {
     _timeZone = [NSTimeZone timeZoneWithName:[resultSet stringForColumn:ReceiptsTable.COLUMN_TIMEZONE]];
     [self setDate:[NSDate dateWithMilliseconds:[resultSet longLongIntForColumn:ReceiptsTable.COLUMN_DATE]]];
     [self setTripName:[resultSet stringForColumn:ReceiptsTable.COLUMN_PARENT]];
+    _customOrderId = [resultSet intForColumn:ReceiptsTable.COLUMN_CUSTOM_ORDER_ID];
+    _categoryId = [resultSet intForColumn:ReceiptsTable.COLUMN_CATEGORY_ID];
```
wbaumann: Or is this why you did it this way?

SmartReceipts/Persistence/DatabaseTableNames.swift
```diff
@@ -19,20 +19,22 @@ enum ReceiptsTable {
...
         static let Category = "category"
+        static let CategoryId = "categoryKey"
         static let Price = "price"
         static let Tax = "tax"
         static let ExchangeRate = "exchange_rate"
         static let Date = "rcpt_date"
         static let Timezone = "timezone"
         static let Comment = "comment"
-        static let Reimbursable = "reimbursable"
+        static let Reimbursable = "expenseable"
```
wbaumann: I forget if I asked, but when did this change apply? When may need to migrate this too if there's an active database out there with `"reimbursable"`

EvsenevDev: I need to check when it was migrated, but `reimbursable` it's not right key for this field.

wbaumann: Worst case we can just include that in our migration

EvsenevDev: Yep, i going to check it and add to migration if needed

PR #207, opened 2018-05-29
Prepared for Release 1.11

EvsenevDev:
- [Fix] Call UIAlertController in Notifications Request Thread
- Added "Initiate Exception" to DEBUG menu
- Commented Custom Exception Handler
- [Fix] Bug in which custom order ids apply in conjunction with older entries

### DIFF
--- a/SmartReceipts/Persistence/Helpers/Database+Receipts.m
+++ b/SmartReceipts/Persistence/Helpers/Database+Receipts.m
@@ -23,7 +23,7 @@
 #import "Constants.h"
 #import "SmartReceipts-Swift.h"
 
-static NSInteger const kDaysToOrderFactor = 1000;
+NSInteger const kDaysToOrderFactor = 1000;
 static NSString * const kGreaterCompare = @" > ";
 static NSString * const kGreaterOrEqualCompare = @" >= ";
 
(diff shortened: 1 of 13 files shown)


SmartReceipts/Services/MigrationService.swift
```diff
@@ -18,26 +18,45 @@ class MigrationService {
...
         for trip in Database.sharedInstance().allTrips() as! [WBTrip] {
             trip.name = WBTextUtils.omitIllegalCharacters(trip.name)
             Database.sharedInstance().update(trip)
         }
     }
     
+    func migrateCustomOrderIds() {
+        for trip in Database.sharedInstance().allTrips() as! [WBTrip] {
+            let receipts = Array(Database.sharedInstance().allReceipts(for: trip).reversed()) as! [WBReceipt]
+            if Database.sharedInstance().deleteReceipts(for: trip) {
```
wbaumann: Can we do an `update` instead of a `delete` and then `save`? I'm worried about it possibly failing after the `delete` operation, causing users to lose their data

SmartReceipts/Persistence/Helpers/Database+Receipts.m
```diff
@@ -294,6 +302,25 @@ - (BOOL)reorderReceipt:(WBReceipt *)receiptOne withReceipt:(WBReceipt *)receiptT
...
+        result = [self receiptsCountInOrderIdGroup:idGroup usingDatabase:db];
+    }];
+    return result;
+}
+
+- (NSInteger)receiptsCountInOrderIdGroup:(NSInteger)idGroup usingDatabase:(FMDatabase *)database {
+    NSString *like = [NSString stringWithFormat:@"'%lu%%'", idGroup];
+    NSArray *components = @[
+                            @"SELECT COUNT(*) FROM ", ReceiptsTable.TABLE_NAME,
+                            @" WHERE ", @"CAST(", ReceiptsTable.COLUMN_CUSTOM_ORDER_ID,  @" AS TEXT) LIKE ", like];
```
wbaumann: Just curious - why the `CAST(custom_order_id) AS TEXT`?

EvsenevDev: @wbaumann it's for check if you have custom order id in order group (days).
For example you have custom order id 17651001, you group id is 17651, so if i want to get count of receipts in this day i need to select receipts where custom order id starts with 17651, i cast to TEXT for check it, i'm no sure if i can do that with INT. I'm not strong in SQLite :( Do you know better way?

wbaumann: Ah that makes sense. Could you add a comment above that `SELECT` statement explaining this above? As part of the next PR is good. Thanks!

EvsenevDev: ok

PR #209, opened 2018-06-08
Some settings polishes

EvsenevDev:
- Layout updates for Settings cells
- [Upgrade] Database version to v17 and add support for payment method custom order ids

### DIFF
--- a/SmartReceipts/Eureka Custom/DescribedButtonCell.swift
+++ b/SmartReceipts/Eureka Custom/DescribedButtonCell.swift
@@ -18,11 +18,15 @@ class DescribedButtonCell: ButtonCellOf<String> {
         textLabel?.isHidden = true
         update()
         
-        title.sizeToFit()
-        subtitle.sizeToFit()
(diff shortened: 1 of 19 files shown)


SmartReceipts/Eureka Custom/DescribedButtonCell.swift
```diff
@@ -18,11 +18,15 @@ class DescribedButtonCell: ButtonCellOf<String> {
         textLabel?.isHidden = true
         update()
         
-        title.sizeToFit()
-        subtitle.sizeToFit()
+        let buttonArrowWidth: CGFloat = 51
```
wbaumann: Why `51`? Or can we make this a `const`?

EvsenevDev: @wbaumann it's for button arrow offset, but, i'm not sure this value is good for constant. I prefer make constants if i use it in some places, but here it's just local value for calculation. But if you think constant is bette, i can make constant. It's just code style. How do you do it on Android?

PR #215, opened 2018-07-12
[Fix] App crashes while rendering a lot of PDF pages

EvsenevDev:


SmartReceipts/PDF/PDFPageRenderView.swift
```diff
@@ -9,39 +9,25 @@
...
 
 class PDFPageRenderView: UIView {
-    var page: CGPDFPage!
     
-    override func draw(_ rect: CGRect) {
-        backgroundColor = UIColor.white
-        
-        let format = UIGraphicsImageRendererFormat.default()
-        // The problem is that we can't use default CGContext for drawing a pdf on iPhone 7 and newer with iOS10 (extended color devices)
-        format.prefersExtendedRange = false
```
wbaumann: Hey quick question for this item: 

There was a bug with iOS10 without this value set, and I know we had fixed some bizarre edge case with it about two years back. Here's the commit:

https://github.com/wbaumann/SmartReceiptsiOS/commit/342581a3560d60f6c87585caed24bbc8688fee50

Is there any issue if we add this stuff back:

```
let format = UIGraphicsImageRendererFormat.default()
// The problem is that we can't use default CGContext for drawing a pdf on iPhone 7 and newer with iOS10 (extended color devices)
format.prefersExtendedRange = false
```

If not, let's do so just to be safe. Not sure if the bug that caused this is still relevant or not, but I'd rather avoid it if possible :)

EvsenevDev: `prefersExtendedRange` is deprecated.
And i rendered on iPhone 8 Plus all works fine, but i don't have iOS 10.
So, we can use it, but not sure we need to use it. What do you think?

wbaumann: Can we add an if/else clause there for if it's iOS 10 or not? Here's the original ticket from a while back:

https://trello.com/c/jfq2APjW/135-bug-fix-an-issue-in-which-pdf-files-do-not-display

wbaumann: Seems it's no longer required, but I just want to avoid adding this bug back for anyone still on iOS 10 (at least until iOS 12 is released and we stop supporting that)

EvsenevDev: @wbaumann what about add it without check iOS version, but add comment like:
`We need to remove it after iOS 12 release`

wbaumann: Sure that works! Just double check that it will works for a test report and then should be good :)

EvsenevDev: @wbaumann done. works fine

wbaumann:
Thanks!

PR #219, opened 2018-07-30
Auto backup 3

EvsenevDev:
- Added Network Listener to SyncService
- Added AppNotificationsCenter - notifications with Rx support
- Added Preferences and Network handlers to SyncService
- Resolved issue double creation Database or Folder files
- Added deletion sync to SyncService

### DIFF
--- a/SmartReceipts/Modules/Backup/BackupView.swift
+++ b/SmartReceipts/Modules/Backup/BackupView.swift
@@ -96,6 +96,7 @@ final class BackupView: UserInterface, GIDSignInUIDelegate {
             }).disposed(by: bag)
         
         wifiSwitch.rx.isOn
+            .skip(1)
             .subscribe(onNext: { [unowned self] in
                 self.presenter.setupUseWifiOnly(enabled: $0)
             }).disposed(by: bag)
--- a/SmartReceipts/Rx/RxSwift+Extensions.swift
+++ b/SmartReceipts/Rx/RxSwift+Extensions.swift
@@ -7,6 +7,7 @@
 //
 
 import RxSwift
+import RxCocoa
 
 extension AnyObserver {
     init(onNext: ((E) -> Swift.Void)? = nil, onError: ((Error) -> Swift.Void)? = nil, onCompleted: (() -> Swift.Void)? = nil) {
@@ -20,5 +21,11 @@ extension AnyObserver {
                 onCompleted?()
             }
         }
+    }
+}
+
+extension Observable {
+    func skipFirst() -> Observable<Element> {
+        return skip(1)
     }
 }
--- a/SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
+++ b/SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
@@ -18,34 +18,46 @@ fileprivate let FOLDER_NAME = "Smart Receipts"
 fileprivate let UDID_PROPERTY = "smart_receipts_id"
 
 class GoogleSyncService: SyncServiceProtocol {
+    
     private let bag = DisposeBag()
     private let syncMetadata = GoogleDriveSyncMetadata()
+    private let syncQueue = DispatchQueue(label: "sync.queue")
     
+    private let dbLock = NSLock()
     func syncDatabase() {
         let dbPath = Database.sharedInstance().pathToDatabase!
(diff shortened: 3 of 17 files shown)


SmartReceipts/Rx/RxSwift+Extensions.swift
```diff
@@ -22,3 +23,9 @@ extension AnyObserver {
         }
     }
 }
+
+extension Observable {
+    func skipFirst() -> Observable<Element> {
+        return skip(1)
```
wbaumann: Should we change the Rx.isOn control (from my comment above) to use this?

SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
```diff
@@ -73,12 +79,45 @@ class GoogleSyncService: SyncServiceProtocol {
...
-        let json = folderCustomProperties()
-        return GoogleDriveService.shared.createFolder(name: FOLDER_NAME, json: json, description: UIDevice.current.name)
-            .do(onSuccess: { [weak self] file in
-                self?.syncMetadata.folderIdentifier = file.identifier
-            })
+    func deleteFile(receipt: WBReceipt) {
+        guard let syncID = receipt.getSyncId(provider: .current), !syncID.isEmpty else { return }
+        GoogleDriveService.shared.deleteFile(id: syncID)
+            .subscribe(onCompleted: {
+                Database.sharedInstance().delete(receipt)
```
wbaumann: Could you add Logger.debug statements to these, so we can trace it more easily? Feel free to do later too

SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
```diff
@@ -73,12 +79,45 @@ class GoogleSyncService: SyncServiceProtocol {
...
+            }, onError: { _ in
+                receipt.isMarkedForDeletion = true
```
wbaumann: I feel like this should be done immediately instead of on error (in case the request times out before it happens). For instance, we can chain this as:

```
{ _ in
receipt.isSynced = false
receipt.isMarkedForDeletion = true
Database.sharedInstance().save(receipt)
}
.flatMap({ _ in 
  GoogleDriveService.shared.deleteFile(id: syncID)
})
.subscribe(onCompleted: {
  Database.sharedInstance().delete(receipt)
})
```

So we always start with the error case as our assumption?

SmartReceipts/Modules/Backup/BackupView.swift
```diff
@@ -96,6 +96,7 @@ final class BackupView: UserInterface, GIDSignInUIDelegate {
             }).disposed(by: bag)
         
         wifiSwitch.rx.isOn
+            .skip(1)
```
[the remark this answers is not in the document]
EvsenevDev: @wbaumann yes

SmartReceipts/Rx/RxSwift+Extensions.swift
EvsenevDev: just if we use rx.isOn.asObservable().skipFirst

SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
EvsenevDev: @wbaumann 
We can't call `save` and `delete` here. I did it for safe managing of tableView. If you remember we had some issues with that and crashes. Which was fixed.
I just can make changes of state flags, and save by error or delete by success. Because our `tableView` of receipts handles our db changes. What do you think about it?

SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
EvsenevDev: @wbaumann sure

SmartReceipts/Services/Auto Backup/GoogleSyncService.swift
wbaumann: @EvsenevDev Ah okay - could you add a comment in there explaining the ordering?

EvsenevDev: @wbaumann yes, i'll comment it and add `Logger.debug`

PR #220, opened 2018-08-03
Auto Backup 4

EvsenevDev:
- Added comments and logs to GoogleDriveService
- Added Take/Retake Image feature
- Added Get Backups feature to Google Drive Backups Provider
- Display this `RemoteBackupMetadata` in list form, including the device 
- Added overflow menu to Backup Item: Restore, Delete, Download to ZIP, [Debug] Download
- Renamed GoogleDriveBackupMetadata to DefaultRemoteBackupMetadata

PR #221, opened 2018-08-05
Auto backups 5

EvsenevDev:
- Changes for PR #220 
- Create]/ a Dialog that presents the user with the option to import a specific backup
- Added an overflow action to "Restore" an existing automatic backup

### DIFF
--- a/SmartReceipts/Modules/Backup/BackupInteractor.swift
+++ b/SmartReceipts/Modules/Backup/BackupInteractor.swift
@@ -9,6 +9,7 @@
 import Foundation
 import Viperit
 import RxSwift
+import Toaster
 
 class BackupInteractor: Interactor {
     let bag = DisposeBag()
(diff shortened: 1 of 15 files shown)


SmartReceipts/Modules/Backup/BackupInteractor.swift
```diff
@@ -24,6 +25,44 @@ class BackupInteractor: Interactor {
         return purchaseService.hasValidSubscriptionValue()
     }
     
+    func importBackup(_ backup: RemoteBackupMetadata, overwrite: Bool) {
+        weak var hud = PendingHUDView.showFullScreen()
+        backupManager?.restoreBackup(remoteBackupMetadata: backup, overwriteExistingData: overwrite)
```
wbaumann: It might be easier to make this process a `Completable`

SmartReceipts/Modules/Backup/BackupInteractor.swift
```diff
@@ -24,6 +25,44 @@ class BackupInteractor: Interactor {
...
+            .subscribe(onCompleted: { [weak self] in
+                guard let safeSelf = self else { return }
+                safeSelf.backupManager?.downloadAllData(remoteBackupMetadata: backup)
+                    .subscribe(onSuccess: { result in
+                        if !result.database.open() { return }
```
wbaumann: And then you can move all of this logic into a `Completable.flatMap` portion, and only use `Toast.show(...)` as the success case. It's the same thing overall, but then it allows all potential errors to propagate to the `onError` block as opposed to be trapped in the subscription.

Not a big concern either way - just a thought!

PR #223, opened 2018-08-11
Auto Backup 7

EvsenevDev:
- Added an overflow action to "Download to ZIP" an existing automatic backup
- Added alert on "Download Zip" error
- Added an overflow action to "[Debug] Download" an existing automatic backup

### DIFF
--- a/SmartReceipts/Modules/Backup/BackupInteractor.swift
+++ b/SmartReceipts/Modules/Backup/BackupInteractor.swift
@@ -65,8 +65,9 @@ class BackupInteractor: Interactor {
             }).filter({ $0 != nil })
             .subscribe(onSuccess: { [weak self] url in
                 self?.presenter.presentOptions(file: url!)
-            }, onError: { error in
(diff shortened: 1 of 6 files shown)


SmartReceipts/Modules/Backup/BackupInteractor.swift
```diff
@@ -29,10 +29,81 @@ class BackupInteractor: Interactor {
         return purchaseService.hasValidSubscriptionValue()
     }
     
+    func downloadZip(_ backup: RemoteBackupMetadata) {
+        weak var hud = PendingHUDView.showFullScreen()
+        backupManager.downloadAllData(remoteBackupMetadata: backup)
+            .map({ result -> URL? in
+                if !result.database.open() { return nil }
```
wbaumann: Maybe have this throw an error instead of `return nil`?

EvsenevDev: @wbaumann good point, i'll change this

PR #225, opened 2018-08-19
Autobackup 8

EvsenevDev:
- [Fix] Bug with renaming receipts after image sync
- When a user clicks the cloud icon, inform him/her about automatic backups
- [Fix] Bug with Upsell Banner
- Import only receipts which not marked for deletion

### DIFF
--- a/SmartReceipts/Modules/Backup/BackupInteractor.swift
+++ b/SmartReceipts/Modules/Backup/BackupInteractor.swift
@@ -134,7 +134,7 @@ class BackupInteractor: Interactor {
                             try? fm.createDirectory(atPath: folder, withIntermediateDirectories: true, attributes: nil)
                         }
                         fm.createFile(atPath: path, contents: file.data, attributes: nil)
-                    }).toArray().map({ _ -> Void in }).asSingle()
+                    }).toArray().asVoid().asSingle()
             }).asCompletable()
             .subscribe(onCompleted: {
                 hud?.hide()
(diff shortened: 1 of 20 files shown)


SmartReceipts/Modules/Backup/BackupInteractor.swift
```diff
@@ -102,31 +102,41 @@ class BackupInteractor: Interactor {
...
-                                try? fm.createDirectory(atPath: folder, withIntermediateDirectories: true, attributes: nil)
-                            }
-                            fm.createFile(atPath: path, contents: file.data, attributes: nil)
-                        }
-                    }
+                    result.append(contentsOf: receipts.filter({ !$0.syncId.isEmpty }))
                 }
                 database.close()
+                return result
+            }).map({ receipts -> [Observable<(WBReceipt, BackupReceiptFile)>] in
```
wbaumann: Let's change this to be a `flatMap` operation (or `flatMapObservable` ). We should only use `map` when going from one object type to another. `flatMap` is an object to a  `Observable` (or `Single`, etc).

EvsenevDev: @wbaumann done, added to next PR with [commit](https://github.com/EvsenevDev/SmartReceiptsiOS/commit/af16f684b6acc84a193688e6b6c4ca39c5a2a2e1#diff-f449d3f1ca6933967ccedb16ad3c3a02L118)

PR #226, opened 2018-09-01
Backup tooltips

EvsenevDev:
- Changes for PR #225 
- Added support SyncErrors
- Added SyncErrors resolving


SmartReceipts/Modules/Backup/BackupInteractor.swift
```diff
@@ -115,27 +110,28 @@ class BackupInteractor: Interactor {
                 }
                 database.close()
                 return result
-            }).map({ receipts -> [Observable<(WBReceipt, BackupReceiptFile)>] in
-                return receipts.map({ [unowned self] receipt in
-                    return self.backupManager.downloadReceiptFile(syncId: receipt.syncId)
+            }).asObservable()
+            .flatMap({ receipts -> Observable<(WBReceipt, BackupReceiptFile)> in
+                return .merge(receipts.map({ receipt in
```
wbaumann: I think this should be `map` instead of `flatMap`, so you can avoid the call to `asObservable` below, right?

SmartReceipts/Modules/Backup/BackupInteractor.swift
```diff
@@ -115,27 +110,28 @@ class BackupInteractor: Interactor {
...
-                        let folder = path.asNSString.deletingLastPathComponent
-                        let fm = FileManager.default
-                        if !fm.fileExists(atPath: folder) {
-                            try? fm.createDirectory(atPath: folder, withIntermediateDirectories: true, attributes: nil)
-                        }
-                        fm.createFile(atPath: path, contents: file.data, attributes: nil)
-                    }).toArray().asVoid().asSingle()
-            }).asCompletable()
+                }))
+            }).map({ downloaded -> Void in
```
wbaumann: Let's make this a `flatMap` again if possible, so it'll properly emit an `onError` if this fails

SmartReceipts/Services/Auto Backup/GoogleDriveService.swift
```diff
@@ -28,6 +30,7 @@ class GoogleDriveService: NSObject, GIDSignInDelegate {
     private override init() {
         gDriveService.shouldFetchNextPages = true
         gDriveService.isRetryEnabled = true
+        spaces = DebugStates.isDebug ? "drive" : "appDataFolder"
```
wbaumann: Maybe we should add a debug toggle for this in the debug panel? Since it'll be useful to test that this works for the `appDataFolder` while in `isDebug == true` mode

EvsenevDev: ok, sure

SmartReceipts/Modules/Backup/BackupInteractor.swift
EvsenevDev: ok

SmartReceipts/Modules/Backup/BackupInteractor.swift
EvsenevDev: i'll try

wbaumann: Let me know if you have any troubles with it. I can walk through a bit more too :)

SmartReceipts/Modules/Backup/BackupInteractor.swift
EvsenevDev: @wbaumann could you explain? :)

wbaumann: Sure thing. So the basic contract is:

* `map`. Takes item object of type `A` and converts it to type `B`
* `flatMap`. Takes item object of type `A` and converts it to type `Observable<B>`

Usually, we want to use a `map` operation when it's a simple conversion between types that cannot fail or have any particular side effects (e.g. mapping a `Receipt` to a `Price`). Since `flatMap` uses an `Observable`, however, it makes more sense to use when we're unsure if there will be success/error conditions. In this case, we're attempting to write a file to disk. What happens if:

* The write fails (e.g. not enough disk space)

In this case, we should emit an `onError` event, but the `map` operation doesn't easily support this. Hence why I thought it would be easier to handle this as a `flatMap` to allow this exception to propagate. Does that make sense?

So one thought would be (psuedo-code below):

```
flatMapIterable(receipts)
     .flatMap(receipt -> 
          // Download Receipt file
          // Emit onError if disk write fails
     )
     .toList()
```

https://github.com/ReactiveX/RxSwift/issues/406

I may also be messing up my RxSwift syntax with the RxJava equivalents, so let me know if I'm giving bad feedback here :)

PR #233, opened 2018-09-16
Updates after review

EvsenevDev:
- [Fix] Debug State - Subscription
- Modify how our backup items are displayed on the backups page
- Google Drive Scope only using the App Data folder in Release
- Modify our backup restoration behavior such that our local database does not track our references to the old sync item
- Sync our `receipts.db`, even when the "Wifi only" setting is on

### DIFF
--- a/SmartReceipts/Services/Auto Backup/SyncService.swift
+++ b/SmartReceipts/Services/Auto Backup/SyncService.swift
@@ -38,6 +38,11 @@ class SyncService {
             }).disposed(by: bag)
     }
     
+    private var canUploadReceipts: Bool {
+        guard let net = network else { return false }
+        return !WBPreferences.autobackupWifiOnly() || net.isReachableOnEthernetOrWiFi
+    }
+    
     func initialize() {
         let center = NotificationCenter.default
         center.addObserver(self, selector: #selector(didInsert(_:)), name: .DatabaseDidInsertModel, object: nil)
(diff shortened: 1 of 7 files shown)


SmartReceipts/Services/Auto Backup/SyncService.swift
```diff
@@ -134,11 +141,13 @@ class SyncService {
         syncProvider = SyncProvider.current
         syncService?.getCriticalSyncErrorStream().bind(to: syncErrorsSubject).disposed(by: bag)
     }
-    
+
     // MARK: - DB Handlers
     
     @objc private func didInsert(_ notification: Notification) {
         syncService?.syncDatabase()
+        if !canUploadReceipts { return }
```
wbaumann: Want to add a `Logger` message to this, stating something like `"Skipping insert due to lack of network availability"`

SmartReceipts/Services/Auto Backup/SyncService.swift
```diff
@@ -150,6 +159,8 @@ class SyncService {
     
     @objc private func didUpdate(_ notification: Notification)  {
         syncService?.syncDatabase()
+        if !canUploadReceipts { return }
```
wbaumann: Same here

SmartReceipts/Services/Auto Backup/SyncService.swift
EvsenevDev: @wbaumann it's not insert skip. Skipping sync images. `didInsert` calls after insert and sync receipts if needed.

wbaumann: Gotcha makes sense

PR #238, opened 2018-10-02
Added support for encrypted secrets

wbaumann:
Adds a simple iOS script, which automatically attempts to decrypt our secrets:

- `SmartReceipts/Service Account.json`. 
- `SmartReceipts/GADConstants.m`. 
- `SmartReceipts/Supporting Files/Secrets.swift`. 

For users that have access to our signing key


EvsenevDev:
@wbaumann nice!

PR #243, opened 2018-10-10
Mofidied our iOS keys to point to the shared Android/iOS keys

wbaumann:


wbaumann:
@EvsenevDev Just wanted to give you a heads up for this change. Let me know if you see any issues with it whenever you have a moment. Thanks!

EvsenevDev:
@wbaumann overall looks good, but some tests are failed.

PR #252, opened 2018-10-16
Applied a variety of translations to the core app experience 

wbaumann:
Screens changed:

- Reports
- Receipts
- Distance
- Tooltips


wbaumann:
@EvsenevDev Build failures aside, do you have any concerns with the changes here? I think they're pretty safe

EvsenevDev:
@wbaumann yes. looks good

PR #253, opened 2018-10-16
Applied translations to our settings page

wbaumann:
This includes the following changes:

1. Applied the majority of our Android translations to the iOS settings page
2. Removed the 'Full Name' setting that is unused
3. Removed the 'CSV Headers' setting, keeping this always on by default
4. Added support for email separation by semi-colon (same as Android)

### DIFF
--- a/SmartReceipts/Modules/Generate Report/GenerateReportShareService.swift
+++ b/SmartReceipts/Modules/Generate Report/GenerateReportShareService.swift
@@ -103,7 +103,7 @@ class GenerateReportShareService: NSObject, MFMailComposeViewControllerDelegate 
     }
     
     fileprivate func split(_ string: String) -> [String] {
-        if (string.contains(",")) {
+        if string.contains(",") {
             // For legacy reasons, we allow these to also be separated by a comma (,)
             return string.components(separatedBy: ",")
         } else {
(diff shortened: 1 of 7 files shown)


wbaumann:
@EvsenevDev Same for this one. All look good? I also made the following changes:

1. Got rid of the "Full Name" setting, since we don't use it anywhere (it was for a one-off white label project)
2. Set the CSV Headers piece to always be `true`, which is the same as Android

I figured it's always easier to delete two extra settings :)

SmartReceipts/Modules/Generate Report/GenerateReportShareService.swift
```diff
@@ -103,7 +103,12 @@ class GenerateReportShareService: NSObject, MFMailComposeViewControllerDelegate
     }
     
     fileprivate func split(_ string: String) -> [String] {
-        return string.components(separatedBy: ",")
+        if (string.contains(",")) {
```
EvsenevDev: Better if you write` if string.contains(",") {` 
Bool expression without (), just codestyle

wbaumann: Ah nice - I'll go ahead and change. Thanks :)

PR #267, opened 2018-10-29
Updated the translations for our iOS app columns

wbaumann:


wbaumann:
@EvsenevDev - Can you take a glance at this one when you're free? I want to see if all looks good to you :)

PR #269, opened 2018-11-03
Refactoring

EvsenevDev:
- [Refactored] Removed useless Objective-C files
- [Refactored] Project Structure for easier navigation
- [Fix] All Warnings


SmartReceipts/Common/UI/AppTheme.swift
```diff
@@ -44,10 +44,6 @@ class AppTheme: NSObject {
         
         // changes color of active elements like buttons and segments, unfortunately alertview cannot be costumized
         UIApplication.shared.keyWindow?.tintColor = primaryColor
-        
-        // chanes status bar color, however iOS built-in controllers (for example mail composer) may override it
-        UIApplication.shared.isStatusBarHidden = false
```
wbaumann: How come we got rid of these? A warning in Xcode right?

SmartReceipts/Modules/Trips/TripsInteractor.swift
```diff
@@ -18,7 +18,7 @@ class TripsInteractor: Interactor {
     
     func configureSubscribers() {
         presenter.tripDeleteSubject.subscribe(onNext: { trip in
-            Logger.debug("Delete Trip: \(trip.name)")
+            Logger.debug("Delete Trip: \(trip.name!)")
```
wbaumann: For a future change, but can we make `name` non-optional? Then we can remove the optional unwinding in these logs (and elsewhere)

SmartReceipts/Common/UI/AppTheme.swift
EvsenevDev: Yes, this is deprecated property

SmartReceipts/Modules/Trips/TripsInteractor.swift
EvsenevDev: @wbaumann yes, we can. But this is Objective-C property by default all looks like optional in Swift. I think we avoid it after rewrite from Objective-C to Swift. And we can obviously mark each property as `nonnull` in Objective-C header.

wbaumann: Ah makes sense. I forgot that was an Obj-C class 👍

PR #270, opened 2018-11-04
Updated many more iOS translations

wbaumann:
Change includes:
 - the removal of unused keys
 - various fixes where our automated script applied the wrong item
 - other places where we moved from Localizable => SharedLocalizable


wbaumann:
I was lazy and didn't run my uni tests, so I'll fix those too :)

EvsenevDev:
> @EvsenevDev Could you check out this one whenever you have a moment? I'll merge in yours first and handle any potential conflicts on my end but just to review :)

Looks good. Where did you found deleted `SettingsViewControllers`? In file system?
I'm not see these files in the project.

wbaumann:
Yeah `SettingsViewControllers` was in the file system. I had a script that looped through and deleted unused strings, but it wasn't removing the ones that were in that class (which is how I found it)

PR #272, opened 2018-11-05
Added a link to our usage in the overflow menu

EvsenevDev:


SmartReceipts/Common/UI/AppTheme.swift
```diff
@@ -11,7 +11,8 @@ import Foundation
 class AppTheme: NSObject {
     
     static let appTitle = "Smart Receipts"
-    
+    static let appTitlePlus = "Smart Receipts Plus"
```
wbaumann: These should probably use `LocalizedString`, but the title is the same in both cases :)

EvsenevDev: @wbaumann could you explain this? It's should be in navigation bar right? You want to localize app title?

wbaumann: Exactly. Right now I think we don't do this (e.g. it's "Smart Receipts" in all locales), but it may be useful for various non-English locales (e.g. Russian) to have a localized title that's more relevant

EvsenevDev: > Exactly. Right now I think we don't do this (e.g. it's "Smart Receipts" in all locales), but it may be useful for various non-English locales (e.g. Russian) to have a localized title that's more relevant

Can we changed it once we need localized here?

wbaumann: Sure thing. I can pick it up as part of my localization ticket. For instance, here's what we have for Russian:

- https://github.com/wbaumann/SmartReceiptsiOS/blob/master/SmartReceipts/Supporting%20Files/ru.lproj/SharedLocalizable.strings#L15

PR #274, opened 2018-11-09
Updated additional string translations

wbaumann:
- We share more resources between the two platforms
- The remaining Localized.strings are better organized with respect to remaining work


SmartReceipts/Modules/Edit Distance/EditDistanceInteractor.swift
```diff
@@ -34,21 +34,12 @@ class EditDistanceInteractor: Interactor {
...
         if database.save(distance) {
             Logger.debug("Distance has been \(asNewDistance ? "updated" : "added")")
-            validateDate(in: distance)
             presenter.close()
         } else {
             Logger.error("Distance can't be \(asNewDistance ? "updated" : "added")")
         }
     }
     
-    private func validateDate(in distance: Distance) {
```
EvsenevDev: Needless validation?

wbaumann: It's probably useful, but I don't have a translated string for that in Android, and this was easier than using Google translate :)

Plus these "Warning: date is outside of report bounds" has confused a lot of users in the past, so I'm slowly working through removing them

PR #277, opened 2018-11-12
Added support for system string translations

wbaumann:
These specifically include the addition of the following Android strings:
 - android.R.string.ok
 - android.R.string.error_message_title
 - android.R.string.untitled

Which have been mapped to our iOS Localization file to support tracking for each.

### DIFF
--- a/SmartReceipts/Models/PaymentMethod.swift
+++ b/SmartReceipts/Models/PaymentMethod.swift
@@ -8,6 +8,9 @@
 
 import FMDB
 
+// For legacy reasons from before we started supporting translations, we use this as the default fallback option
+fileprivate let LEGACY_DEFAULT_PAYMENT_METHOD = "Unspecified"
+
 @objcMembers
 class PaymentMethod: NSObject, FetchedModel, Pickable {
     var objectId: UInt = 0
@@ -53,8 +56,7 @@ class PaymentMethod: NSObject, FetchedModel, Pickable {
         for method in allMethods! {
             if method.method == LocalizedString("payment_method_unspecified", comment: "") {
                 return method
-            } else if method.method == "Unspecified" {
-                // For legacy reasons from before we started supporting translations, we use this as the default fallback option
+            } else if method.method == LEGACY_DEFAULT_PAYMENT_METHOD {
                 return method
             }
         }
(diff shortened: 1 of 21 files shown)


SmartReceipts/Models/PaymentMethod.swift
```diff
@@ -51,7 +51,10 @@ class PaymentMethod: NSObject, FetchedModel, Pickable {
     class func defaultMethod(_ database: Database = Database.sharedInstance()) -> PaymentMethod {
         let allMethods = database.allPaymentMethods()
         for method in allMethods! {
-            if method.method == LocalizedString("payment.method.unspecified", comment: "") {
+            if method.method == LocalizedString("payment_method_unspecified", comment: "") {
+                return method
+            } else if method.method == "Unspecified" {
+                // For legacy reasons from before we started supporting translations, we use this as the default fallback option
```
wbaumann: This is probably the main thing worth looking at. Let me know if you think of a better approach for this

EvsenevDev: It's ok, but what about adding it to constants? Or it's not makes sense?

SmartReceipts/Supporting Files/pt.lproj/Localizable.strings
```diff
@@ -1,3 +1,9 @@
+/** Android System Strings (ok, untitled, error_message_title) **/
+"generic_button_title_ok" = "OK";
+"generic_error_alert_title" = "Erro";
+"payment_method_unspecified" = "<Sem título>;
```
EvsenevDev: Can we add `"` before semicolon, looks like this is not one translation without `"`. This is cannot be builded

wbaumann: Oops my fault. I'll fix this

SmartReceipts/Models/PaymentMethod.swift
wbaumann: Sure - I'll add it as a constant to this file

wbaumann:
Note: Don't review yet. I still need to fix some failing tests :)

wbaumann:
@EvsenevDev All should be good for this one whenever you're free. No rush

PR #278, opened 2018-11-13
Updates

EvsenevDev:
- Changed support email
- [Fix] Bug in which spaces don't appear when typing
- Added support for sharing a single receipt image

### DIFF
--- a/SmartReceipts/Common/Constants/Constants.m
+++ b/SmartReceipts/Common/Constants/Constants.m
@@ -30,6 +30,6 @@ void SRDelayedExecution(NSTimeInterval seconds, ActionBlock action) {
     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, seconds * NSEC_PER_SEC), dispatch_get_main_queue(), action);
 }
 
-NSString *const FeedbackEmailAddress =              @"supp" @"port" @"@sma" @"rtreceipts.co";
+NSString *const FeedbackEmailAddress =              @"sup" @"port" @"@sma" @"rtreceipts.co";
 NSString *const FeedbackEmailSubject =              @"Smart Receipts - Feedback";
 NSString *const FeedbackBugreportEmailSubject =     @"Smart Receipts - Support";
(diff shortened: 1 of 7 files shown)


SmartReceipts/Common/Constants/Constants.m
```diff
@@ -30,6 +30,6 @@ void SRDelayedExecution(NSTimeInterval seconds, ActionBlock action) {
     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, seconds * NSEC_PER_SEC), dispatch_get_main_queue(), action);
 }
 
-NSString *const FeedbackEmailAddress =              @"will.r.baum" @"ann@gm" @"ail.com";
+NSString *const FeedbackEmailAddress =              @"supp" @"port" @"@sma" @"rtreceipts.co";
```
wbaumann: I think this is misspelled. It's currently

- suppport@smartreceipts.co

But should be:

- - support@smartreceipts.co

One too many 'p' letters :)

SmartReceipts/Common/Utils/EurekaWhitespaceWorkaround.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import Foundation
+import Eureka
+
+final class EurekaWhitespaceWorkaround {
+    
+    static func configureTextCells() {
+        let workaround = EurekaWhitespaceWorkaround()
+        TextRow.defaultCellSetup = { cell, row in
+            // the workaround object is retained here on purpose
```
wbaumann: Want to add another comment, which points to that issue link? Thanks!

SmartReceipts/Common/Utils/EurekaWhitespaceWorkaround.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  EurekaWhitespaceWorkaround.swift
+//  SmartReceipts
+//
+//  Created by Bogdan Evsenev on 13/11/2018.
+//  Copyright © 2018 Will Baumann. All rights reserved.
```
wbaumann: Or up here for the link to the Eureka issue, just so we can track here

SmartReceipts/Common/Constants/Constants.m
EvsenevDev: @wbaumann yes, sorry 👍

SmartReceipts/Common/Utils/EurekaWhitespaceWorkaround.swift
EvsenevDev: @wbaumann makes sense, i'll add it to this line

PR #280, opened 2018-11-15
Added a fallback language support to English

wbaumann:
By default, iOS will return the key value rather than a defined translation. To simplify things for our users, I've added support for a default fallback mechansim to the english bundle.

### DIFF
--- a/SmartReceipts/Common/Constants/Global.swift
+++ b/SmartReceipts/Common/Constants/Global.swift
@@ -33,9 +33,24 @@ func delayedExecution(_ afterSecons: TimeInterval, closure: @escaping () -> ()) 
 }
 
 func LocalizedString(_ key: String, comment: String = "") -> String {
-    var result = NSLocalizedString(key, tableName: nil, comment: comment)
+    // By default, attempt to load from our Shared set of strings
+    var result = NSLocalizedString(key, tableName: "SharedLocalizable", comment: comment)
     if result == key {
-        result = NSLocalizedString(key, tableName: "SharedLocalizable", comment: comment)
+        // If we failed to find this string in our SharedLocalizable.strings file, check Localizable.strings one
+        result = NSLocalizedString(key, tableName: nil, comment: comment)
+    }
+    if result == key {
+        // If we cannot find it in either, fall back to English
+        if let path = Bundle.main.path(forResource: "en", ofType: "lproj") {
+            if let enBundle = Bundle(path: path) {
+                // Check the English Localizable.strings file
+                result = NSLocalizedString(key, bundle: enBundle, comment: comment)
+                if result == key {
+                    // And finally fall back to the English SharedLocalizable.strings file
+                    result = NSLocalizedString(key, tableName: "SharedLocalizable", bundle: enBundle, comment: comment)
+                }
+            }
+        }
     }
     return result
 }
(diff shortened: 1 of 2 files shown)


SmartReceipts/Common/Constants/Global.swift
```diff
@@ -33,9 +33,24 @@ func delayedExecution(_ afterSecons: TimeInterval, closure: @escaping () -> ())
...
+    var result = NSLocalizedString(key, tableName: "SharedLocalizable", comment: comment)
     if result == key {
-        result = NSLocalizedString(key, tableName: "SharedLocalizable", comment: comment)
+        // If we failed to find this string in our SharedLocalizable.strings file, check Localizable.strings one
+        result = NSLocalizedString(key, tableName: nil, comment: comment)
+    }
+    if result == key {
+        // If we cannot find it in either, fall back to English
+        if let path = Bundle.main.path(forResource: "en", ofType: "lproj") {
+            if let enBundle = Bundle(path: path) {
```
wbaumann: One thought - should we cache this bundle in memory via a `lazy`?

EvsenevDev: @wbaumann not sure if we should, but i think it will be better

SmartReceipts/Common/Localization/LocalizedString.m
```diff
@@ -11,9 +11,26 @@
...
+    // By default, attempt to load from our Shared set of strings
+    NSString* result = NSLocalizedStringFromTable(key, @"SharedLocalizable", comment);
     if ([result isEqualToString:key]) {
-        result = NSLocalizedStringFromTable(key, @"SharedLocalizable", comment);
+        // If we failed to find this string in our SharedLocalizable.strings file, check Localizable.strings one
+        result = NSLocalizedString(key, comment);
+    }
+    if ([result isEqualToString:key]) {
+        // If we cannot find it in either, fall back to English
+        NSString* path = [[NSBundle mainBundle] pathForResource:@"en" ofType:@"lproj"];
```
EvsenevDev: How it should works? We define app supports Spanish for example, but we can't found string for key and you want to show English, right? How we can found missing translations? What about add logs for that?

SmartReceipts/Common/Constants/Global.swift
wbaumann: Sounds good. I'll go ahead and make this change

SmartReceipts/Common/Localization/LocalizedString.m
wbaumann: Good point about:

> How we can found missing translations? What about to add logs for that?

Maybe I'll disable this fallback in `DEBUG` mode, so we can see easily? I'll add logs too. I just think it's better to have some message there than nothing

EvsenevDev: Ok, thanks!

SmartReceipts/Common/Constants/Global.swift
wbaumann: Actually from the `Bundle` source:

> Because NSBundle caches allocated instances, subclasses should be prepared to receive an already initialized object back from [super initWithPath:]

So I think we're okay without caching this with a `lazy`

SmartReceipts/Common/Localization/LocalizedString.m
wbaumann: @EvsenevDev I added a debug log statement here. Let me know if you think we should do anything else. No rush of course :)

EvsenevDev: @wbaumann Merged 👍

SmartReceipts/Common/Constants/Global.swift
EvsenevDev: Ok. Thanks!

PR #281, opened 2018-11-16
Removed an iOS settings that lacks an Android equivalent

wbaumann:
Rather than supporting the ability to allow or disallow showing receipts out of date bounds, matching the behavior that we have on Android.

### DIFF
--- a/SmartReceipts/Modules/Edit Receipt/EditReceiptInteractor.swift
+++ b/SmartReceipts/Modules/Edit Receipt/EditReceiptInteractor.swift
@@ -111,7 +111,6 @@ class EditReceiptInteractor: Interactor {
     
     private func validateDate(in receipt: WBReceipt) {
         Observable<Void>.just(())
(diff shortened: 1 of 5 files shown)


wbaumann:
I'm going to close this and fix the merge conflicts in another PR. Updates to follow :)

SmartReceipts/Modules/Edit Receipt/EditReceiptInteractor.swift
```diff
@@ -111,7 +111,6 @@ class EditReceiptInteractor: Interactor {
     
     private func validateDate(in receipt: WBReceipt) {
         Observable<Void>.just(())
-            .filter({ !WBPreferences.allowDataEntryOutsideTripBounds() })
```
EvsenevDev: Always disallowed now?

wbaumann: It should just always show now

PR #282, opened 2018-11-16
Finalized all iOS string keys for translation

wbaumann:
This includes the following changes:

- Formatting all string keys to use underscores for spaces instead of periods
- Replaced a few unique error titles with the "Error" generic one
- Deleted a few strings that could be replaced by Shared equivalents

Following this change, all remaining items can be directly translated into their respective languages


wbaumann:
Closing this one in favor of:

- https://github.com/wbaumann/SmartReceiptsiOS/pull/284

Which resolves the conflicts

PR #299, opened 2018-12-28
Added Quick Actions feature for Receipts

EvsenevDev:


SmartReceipts/Services/Local/QuickActionService.swift
```diff
@@ -0,0 +1,118 @@
...
+    private var subscription: Disposable?
+    private weak var view: UIViewController!
+    private let scanService = ScanService()
+    
+    init(view: UIViewController) {
+        self.view = view
+    }
+    
+    func configureQuickActions() {
+        UIApplication.shared.shortcutItems = lastOpenedTrip == nil ?
```
wbaumann: Does this run on the UI Thread? Any risks for calling `Database.sharedInstance().tripWithName(tripName)` from `lastOpenedTrip` here that could cause it to freeze?

EvsenevDev: yes it's in Main Thread, but we call `configureQuickActions` on app termination. If we call it in background app will terminate without configuration.

wbaumann: Ah - so this only happens when the app is closing? That's okay then. I just didn't want to block on launch then. Thanks!

PR #309, opened 2019-02-06
Fix crashes

EvsenevDev:


wbaumann:
@EvsenevDev Thanks so much! Could you also send this over to `master`? There were some merge conflicts when I tried to `cherry-pick` this over

PR #314, opened 2019-02-23
Fixed Retain Cycles and Memory Leaks

EvsenevDev:
- [Fix] Memory Leak in Edit Receipt Module
- [Fix] Retain Cycles in some screens
- [Fix] fixed results from BackupFilePicker


wbaumann:
Hey - is it worth cherry-picking any of these into our current release in your view?

EvsenevDev:
> Hey - is it worth cherry-picking any of these into our current release in your view?

I hope these changes should fix the crash and it fixed memory leaks. So, i think yes, you can cherry-pick it into current release.

PR #320, opened 2019-04-22
Hotfixes for 1.14.5

EvsenevDev:
- [Fix] DB migration with French locale - v11 string escaping issue
- Added LSSupportsOpeningDocumentsInPlace = YES by Apple Requirement


wbaumann:
@EvsenevDev Overriding this merge, since it seems to be an issue with Travis and the newer Swift syntax. Let's fix this with the next PR. Thanks!

PR #326, opened 2019-07-22
Dev 1.16(2)

EvsenevDev:
- Fixed Base Currency Price for Edit Mode
- Added Weekday to Date inline rows
- [Fix] Generate ZIP for Report without files
- [Fix] NavigationBar on iPad (iOS 10.3)
- [Fix] show Autocomptions by focus on predictiveCell
- [Fix] issue in which the date's always default to the trip start date


wbaumann:
Thanks @EvsenevDev !

PR #328, opened 2019-08-22
Currency calculates fixes

EvsenevDev:
- [Fix] few bugs with the currency exchange rate calculations for 1.16
- [Fix] Item #3 with recalculate Base Currency Rate


wbaumann:
@EvsenevDev If you have a moment, would you be able to cherry-pick this into 1.16, bump the version number, and upload a release?

PR #340, opened 2020-01-12
New UI/UX

EvsenevDev:
Most part of a big re-design of app.


SmartReceipts/Common/UI/ActionSheet/ActionSheet.swift
```diff
@@ -0,0 +1,208 @@
...
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        titleLabel?.sizeToFit()
+        titleLabel?.frame.origin.x = UI_MARGIN_16
+        
+        titleLabel?.center.y = bounds.midY
+        
+        guard let imageView = imageView else { return }
+        imageView.center.y = bounds.midY
+        imageView.frame.origin.x = bounds.width - imageView.bounds.width - 19
```
wbaumann: I always worry a bit about magic values here like `19`, but not a blocker. In the future, I think it's good to make these constants, so it's clearer to read what their intent is :)

EvsenevDev: Me too, seems like i tested and forgot to move it in private constants

PR #345, opened 2020-02-18
Update 1.18.5

EvsenevDev:
- Fixed crash with NSNotFound while deleting receipt
- Added Date format options
- Updated UI for Tooltip


wbaumann:
@EvsenevDev - Looks like the build failed... Not sure if a Travis issue?

PR #348, opened 2020-03-29
Tooltips update

EvsenevDev:
- Updated logic for Privacy tooltip
- Added tooltip with Plus users and backup recovery
- Added tooltip to explain how Smart Receipts reports/trips are suppose to function
- Added tooltip with PaymentMethod question
- Insert PaymentMethod by tap YES in tooltip


wbaumann:
@EvsenevDev - Looks good but now the CircleCI build is failing (I know you deleted this). Do you know where we can remove it from our GitLab build script?

